PR #11, opened 2023-06-19
Simplify interface.gd

ajreckof:
This is a pass to simplify interface.gd as there was multiple function doing the same thing. Moreover it moved intinialisation of value to the Init of tag by creating a new optional argument that if he is not null change initial value of the property.


MewPurPur:
Thanks!

PR #27, opened 2023-09-23
Implement a save system and handles

MewPurPur:
Implements a save system and settings. For now they are extremely barebones, the only option is to remember your window mode when you exit, but they should be easily expandable.

Implements control handles. As it stands, they only exist for <circle> and <ellipse>, and they are pretty much completely broken. I'll look into some refactoring work to allow handles to nicely correspond to number fields.


MewPurPur:
I will merge this for now and work on the refactoring that would allow to fix the bug later.

PR #79, opened 2023-10-14
Rework PathCommandArray into AttributePath

### LINKED ISSUE #53: Path handles can't be dragged properly
Editing path data causes everything to rebuild.

This is very challenging. Paths have several ways they can be edited.

- LineEdit changed or outside changes -> Change `value` -> Generate new path commands -> Rebuild inspector without autogenerating new `value`
- Commands added or removed via UI -> Change path commands -> Rebuild inspector + Autogenerate new `value`
- Values in a command changed via UI -> Change a single path command ->  Autogenerate new `value` without rebuilding the inspector

But...

- `value` has a setter that propagates the change to other parts of the codebase. It's also the thing that can be changed by other parts of the codebase.
-  So changing value should always update the inspector.
- Yet, updating a single command via UI should update the value, but not the inspector.

How do I get out of this conundrum? I don't know, but it's a very high priority bugfix.

### LINKED ISSUE #70: A/C path commands can offset the whole editor
image

Arcs and Curves are just too wide, when they stretch wider than the inspector, they can mess up the UI.

Yet, I feel like what needs to be done here is come up with a smaller widget for editing XY pairs of values. (#54)

MewPurPur:
Reworks PathCommandArray into an AttributePath that extends Attribute. PathCommand is in its own class now also. The reason for this is that there is a default list of commands, which should reduce the need to parsing paths multiple time.

Fixes #53. This might also help with performance, although I'm not sure it will have much of an effect.

Also refactors NumberField to not use arrows, as they just didn't seem all that useful.  The PR some visuals, makes it so number field text sticks out when it's different from the default value.

Sliders are implemented in number fields to replace arrows,but they are only optionally used when the number field has a min and max. Number fields are now a bit bigger by default when without a slider, but still smaller than before (Fixes #70)

image

Fixes certain path parsing bugs too.


MewPurPur:
HOLY SHIT I ACCIDENTALLY FIXED THE BIGGEST BUG

Issue #58, opened 2023-10-10
Improve the Import dialog


MewPurPur:
It's an abomination. I should look into whether it can just use the default OS file manager.

If not, it still needs better theming - but it's really hard to work with the Godot node, so I hope I don't have to fix everything myself.

MewPurPur:
I'll close this when I improve the theming of the dialog a little more.

Kiryonn:
You know it's just a checkbox away right ?
![image](https://github.com/MewPurPur/GodSVG/assets/60699390/8963f948-d605-4751-aca3-badf0dd39ccf)

Kiryonn:
it's still missing on the webbuild. I haven't checked the software build yet

Kiisu-Master:
I added importing with file dialog to web build in #515!
It just isn't in the online web-build of GodSVG because we're waiting for Godot to release 4.2.2 stable.

Kiryonn:
thanks

PR #86, opened 2023-10-16
Implement native File Dialog with fallback

Swarkin:
"Fixes" #58
Implement a native file Dialog using `DisplayServer.file_dialog_show` while keeping the old functionality as fallback.


MewPurPur:
Thanks!

Issue #55, opened 2023-10-10
Implement Undo and Redo


MewPurPur:
Use Godot's UndoRedo to make actions reversible. Add `Ctrl+Z` as a shortcut; for Redo, allow both `Ctrl+Y` and `Ctrl+Shift+Z`. For an initial implementation, the whole SVG's state can be saved after each action, rebuilding all global definitions and tag editors in the process.

Interestingly, the code editor's own UndoRedo already works quite decently. But it considers every adjustment as an action, even if you're just holding a spinner arrow or dragging a handle.

MewPurPur:
Not working on this _right_ now, turned out to be a tougher nut to crack than I hoped.

SeremTitus:
#104

MewPurPur:
Pinning as solving this important for the first alpha.

PR #104, opened 2023-10-22
UndoRedo support

SeremTitus:
Using built-in class `UndoRedo`  add undo redo functions to the project.

Input mapping is maintained as common in other programs


SeremTitus:
#55

MewPurPur:
Thanks! I'll wait for you to mark it as non-WIP. By the way, I'm absolutely cool with a cheaper approach like just saving the string after each action.

Also, how do you want to be listed in the author list? Formats would be "Real Name", "pseudonym", or "Real Name (pseudonym)"?

SeremTitus:
> Thanks! I'll wait for you to mark it as non-WIP. By the way, I'm absolutely cool with a cheaper approach like just saving the string after each action.
> 
> Also, how do you want to be listed in the author list? The format can be "Real Name", "pseudonym", or "Real Name (pseudonym)"?

I will continue working and testing this branch and I will mark it when every thing is done. Concerning my name  I  prefer `Serem Titus` , Which are my real names

MewPurPur:
Actually, rather not have only the real name, so things can be tracked a little more easily to a Github username, and also so pseudonyms are distinct from real names. So would "Serem Titus (SeremTitus)" be good?

SeremTitus:
> Actually, rather not have only the real name, so authors can be tracked a little more easily to their contributions, and also so pseudonyms are distinct from real names. So would "Serem Titus (SeremTitus)" be good? Sorry about the earlier miscommunication

That will be okay

MewPurPur:
Thanks a ton for trying to tackle this! I've tested the PR and looked at the code:

**Code:**

- Very unsure with the technical complexity of this implementation and how much more difficult it would make it to implement new features. I insist that it might be a better idea to save the SVG text at the right times, instead of the resources GodSVG works with. That'd be much nicer for future contributors, they'd only need to know when in code to ask for UndoRedo to be saved.
- UndoRedo should not be tied into GodSVG's interface in any way! Namely, it should not be managing input fields or handles. Input events and handles should be able to modify the underlying resources (Attribute, Tag, etc.), and to listen to changes in them so that they can update. There should NOT be a third party able to modify them, namely an UndoRedo.

**Testing:**

- I tried moving a handle around and undoing it; the movements were saved in the undoredo frame-by-frame. Moreover, the X and Y movements were saved as separate actions. Since the attributes are modified frame-by-frame, we'd need some way to mark when you drop the handle and record only that action as a change for UndoRedo to act on. The logic for this is in handles_manager.gd and the Handles classes definition.
- **I'll implement bulk attribute changing to address the X and Y being separate actions issue.**
- After undoing all the way back, I got a crash in `input_field_undo_redo_action()`. I get this same crash in a lot of circumstances. (though as discussed, UndoRedo shouldn't be handling these in the first place)
- Dragging around the slider of number fields has the same problem with registering every frame.
- I saw inconsistent behavior in redoing, where sometimes the order of actions would be messed up. For example, I had a minified path (so changing a parameter changes it back to its default spacing). I changed a path parameter and tried undoing and redoing it. Undo had two steps: Undoing the parameter change, and minifying the path back. Redoing did them in the same order instead of the opposite, so in the end I came back with a unminified path with the old parameters, instead of the new ones.

MewPurPur:
Careful, I also fixed the rect edit connections in another PR although it doesn't list it as a conflict here.

SeremTitus:
> Thanks a ton for trying to tackle this! I've tested the PR and looked at the code:
> 
> **Code:**
> 
> * Very unsure with the technical complexity of this implementation and how much more difficult it would make it to implement new features. I insist that it might be a better idea to save the SVG text at the right times, instead of the resources GodSVG works with. That'd be much nicer for future contributors, they'd only need to know when in code to ask for UndoRedo to be saved.
> * UndoRedo should not be tied into GodSVG's interface in any way! Namely, it should not be managing input fields or handles. Input fields and handles should be able to modify the underlying resources (Attribute, Tag, etc.), and to listen to changes in them so that they can update. There should NOT be a third party able to modify them.
> 
> **Testing:**
> 
> * I tried moving a handle around and undoing it; the movements were saved in the undoredo frame-by-frame. Moreover, the X and Y movements were saved as separate actions. Since the attributes are modified frame-by-frame, we'd need some way to mark when you drop the handle and record only that action as a change for UndoRedo to act on. The logic for this is in handles_manager.gd and the Handles classes definition.
> * ~**I'll implement bulk attribute changing to address the X and Y being separate actions issue.**~ Done!
> * After undoing all the way back, I got a crash in `input_field_undo_redo_action()`. I get this same crash in a lot of circumstances. (though as discussed, UndoRedo shouldn't be handling these in the first place)
> * Dragging around the slider of number fields has the same problem with registering every frame.
> * I saw inconsistent behavior in redoing, where sometimes the order of actions would be messed up. For example, I had a minified path (so changing a parameter changes it back to its default spacing). I changed a path parameter and tried undoing and redoing it. Undo had two steps: Undoing the parameter change, and minifying the path back. Redoing did them in the same order instead of the opposite, so in the end I came back with a unminified path with the old parameters, instead of the new ones.

I did not see this earlier I have not checked Github for a while, the initial attempt was proving to cause regression with every new commit so I decided to rewrite it  with better understanding of the limitations... In the new pull request  #116  the implementation is more passive and fail proof

PR #115, opened 2023-10-24
Fix UI scaling weirdly

Kiisu-Master:
I tried to fix the UI scaling in weird ways.
I am sure this is not the best way of fixing it, because I did some hacky things and didn't look into how it all actually works.
Sorry

Current problems with this fix: 
- The viewport is not scaling properly and jumps when moving the split. (Because the size gets rounded I think)
- The window minimum size should probably be set somewhere else. (Not in the settings loading)
- The code editor goes to minimum size when typing text in it and jumps back to full size when moving any splits (Maybe a Godot bug)

Screenshot of what it looks like: 
image


MewPurPur:
Thanks! I will look into this soon and provide feedback.

Kiisu-Master:
Just Kiisu-Master, or Kiisu_Master which is my usual username which I couldn't use on github.

Also I found that the scaling was set like that for high res screens  ( #4  ).
On my Laptop (with a low res screen) it made the whole UI pixelated/smooth and too big, so that's why I changed it.
Maybe there is a better way to support high res screens?

Kiisu-Master:
Maybe a user setting that changes `get_window().content_scale_factor`

MewPurPur:
There are a lot of things that must be addressed after this, but I think it's the right path forward for now. A `content_scale` setting providing an enum would also be nice to have, but reeeaaally complex to implement. It would need to modify the theme resource and make all the little editors respond well to `THEME_CHANGED` notifications. Are you up to the task?

For now, please try to address the following issues:

- Snapping lines aren't drawn over the resizing grabber.

image

- `zoom_reset()` doesn't fit the image very well to frame when the viewport becomes bigger.

- The checkerboard applied is too small when the viewport becomes large, so this should be taken into account too when passing it to the checkerboard shader.

- Rounding issue would be nice to be fixed too, yeah.

- The max inspector size needs to be toned down quite a bit, it's pretty ridiculous right now.

MewPurPur:
May I request splitting off the horizontal grabber in another PR, so I can merge that enhancement separately?

Kiisu-Master:
> May I request splitting off the horizontal grabber in another PR, so I can merge that enhancement separately first?

Why do you want to merge that separately?
The split between inspector and viewport doesn't work properly yet because the viewport doesn't scale.
It's almost the only thing in this PR anyways.

MewPurPur:
The ability to adjust the proportion between the inspector and display is separate from changing the stretch mode, they don't require each other to make sense.

MewPurPur:
I talked about this approach and why I can't accept a PR solely implementing it in #342

MewPurPur:
YOLO

PR #137, opened 2023-10-27
improve update_root_tag

SeremTitus:
Takes updates from code editor and applyes them to there existing attributes rather than replacing child_tags

### DIFF
--- a/src/SVG.gd
+++ b/src/SVG.gd
@@ -15,7 +15,7 @@ func _ready() -> void:
 	
 	if GlobalSettings.save_svg:
 		string = GlobalSettings.save_data.svg
-		sync_data()
+		string_to_replace_root_tag(string)
 	else:
 		update_string()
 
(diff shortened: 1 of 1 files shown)


MewPurPur:
Please add comments to explain what the method does and how it does it.

**Testing:**

Importing SVGs is just broken. It seems like it only can add a single tag, and it also crashes if the SVG had no tags beforehand.

Also, since latest main branch has big changes to how SVGs are imported, please rebase to that.

SeremTitus:
> Please add comments to explain what the method does and how it does it.
> 
> **Testing:**
> 
> Importing SVGs is just broken. It seems like it only can add a single tag, and it also crashes if the SVG had no tags beforehand.
> 
> Also, since latest main branch has big changes to how SVGs are imported, please rebase to that.

Okay , I will try my best to test it better and comment how it works

SeremTitus:
I think I was wrong about this function it is extremely buggy not worth implimenting

PR #154, opened 2023-10-30
Fix crash when deleting tag

Kiisu-Master:
The delete function had changed name and usage, but this wasn't updated


MewPurPur:
Thanks!

PR #162, opened 2023-11-02
Add Invert zoom setting #119

### LINKED ISSUE #119: Implement a setting to switch zoom direction
People have different preferences between zooming up/down to zoom in/out. I think we should accommodate that with a setting in GlobalSettings that's checked against when using the mouse scrollwheel.

Kiisu-Master:
Closes #119 
_Production edit: Supersedes #125_

~This is still missing translation strings.~
~I also haven't tested with latest version from main.~ it works


MewPurPur:
For clarity, you only need to worry about EN localization. I'll be supporting the BG one.

src/ui_parts/viewport.gd
```diff
@@ -37,12 +37,18 @@ func clamp_view() -> void:
...
+		zoom_level *= sqrt(2)
 	center_frame()
 
-func zoom_out() -> void:
-	zoom_level /= sqrt(2)
+func zoom_out(pressed_button: bool) -> void:
+	if GlobalSettings.invert_zoom and not pressed_button:
+		zoom_level *= sqrt(2)
+	else:
+		zoom_level /= sqrt(2)
```
MewPurPur, line 51: This is not where it should be done, instead, GodSVG should use the invert_zoom value to decide which one to call in the first place.

Kiisu-Master: so would this be better?
```gd
		match event.button_index:
			MOUSE_BUTTON_WHEEL_UP:
				if GlobalSettings.invert_zoom:
					zoom_out()
				else:
					zoom_in()
			MOUSE_BUTTON_WHEEL_DOWN:
				if GlobalSettings.invert_zoom:
					zoom_in()
				else:
					zoom_out()
```

MewPurPur: Yes.

MewPurPur:
Thanks!

PR #170, opened 2023-11-02
Add German translation

Swarkin:
Adding German translation to GodSVG.

A couple things to note:
- I wasn't sure if i should change the word "Tag" to something else because I couldn't find a similar and short word for it in German. Problem is, "Tag" means "day" and could be confusing.
- Same applies to "Handles", couldn't find a similar word.
- Is the word SVG feminine, masculine or neutral in German? Which articles should be used? I assumed it to be masculine.
- All other translated strings should be good to go.

Please tell me whether those translations are alright (preferably by someone whose main language is German as well).

### DIFF
--- a/src/GlobalSettings.gd
+++ b/src/GlobalSettings.gd
@@ -13,6 +13,9 @@ const default_config = {
 		"save_svg": true,
 		"save_window_mode": true,
 	},
+	"input": {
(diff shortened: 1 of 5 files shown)


MewPurPur:
Make sure to rebase this since #162 was merged.

PR #182, opened 2023-11-05
Rework ZoomMenu

thiagola92:
This would rework `ZoomMenu` to be a scene with it own logic, removing a little of logic from `Viewport` and `Display`.  

Changes:
- `ZoomMenu` would get full control over `zoom_level` and buttons to control it.
- `ZoomMenu` will notify others about zoom change with signals.
- Change `display`, `view`, `controls`, `display_texture` to use `@export`.
  - Using `@onready` creates a need to use `await get_tree().process_frame` (to wait variables be setted). As they exists from the start, there is no reason to not use `@export` and point at them.
- Refactor match condition in `Viewport` to use `when`.
  - This could cause problems because is quite new.

**Note**: I'm just proposing, it's okay if doesn't fit this repository style.

**Note 2**: For some reason I always have problem when opening GodSVG project the first time right after cloning. But opening a second time seems to correct :thinking:

### DIFF
--- a/src/ui_parts/display/menu/zoom_menu/zoom_menu.gd
+++ b/src/ui_parts/display/menu/zoom_menu/zoom_menu.gd
@@ -2,12 +2,14 @@ extends HBoxContainer
 
 
 const MIN_ZOOM = 0.125
-
 const MAX_ZOOM = 64.0
 
 signal zoom_changed(zoom_level: float)
+signal zoom_reset_pressed()
 
-signal zoom_reseted()
+@onready var zoom_out_button: Button = $ZoomOut
+@onready var zoom_in_button: Button = $ZoomIn
+@onready var zoom_reset_button: Button = $ZoomReset
 
 var zoom_level: float:
 	set(value):
@@ -24,6 +26,8 @@ func _ready() -> void:
 func zoom_out() -> void:
 	zoom_level /= sqrt(2)
 
+func zoom_in() -> void:
+	zoom_level *= sqrt(2)
 
 # Choose an appropriate zoom level and center the camera.
 func zoom_reset() -> void:
@@ -32,23 +36,19 @@ func zoom_reset() -> void:
 	zoom_level = float(nearest_po2(int(8192 / maxf(svg_attribs.width.get_value(),
 			svg_attribs.height.get_value()))) / 32.0)
 	
-	zoom_reseted.emit()
-
-
-func zoom_in() -> void:
-	zoom_level *= sqrt(2)
+	zoom_reset_pressed.emit()
 
 
 func _update_buttons_appearance() -> void:
-	%ZoomReset.text = String.num(zoom_level * 100,
+	zoom_reset_button.text = String.num(zoom_level * 100,
 			2 if zoom_level < 0.1 else 1 if zoom_level < 10.0 else 0) + "%"
 	
-	%ZoomIn.disabled = zoom_level >= MAX_ZOOM
-	%ZoomIn.mouse_default_cursor_shape = (
-		Control.CURSOR_ARROW if %ZoomIn.disabled else Control.CURSOR_POINTING_HAND
+	zoom_in_button.disabled = zoom_level >= MAX_ZOOM
+	zoom_in_button.mouse_default_cursor_shape = (
+		Control.CURSOR_ARROW if zoom_in_button.disabled else Control.CURSOR_POINTING_HAND
 	)
 	
-	%ZoomOut.disabled = zoom_level <= MIN_ZOOM
-	%ZoomOut.mouse_default_cursor_shape = (
-		Control.CURSOR_ARROW if %ZoomOut.disabled else Control.CURSOR_POINTING_HAND
+	zoom_out_button.disabled = zoom_level <= MIN_ZOOM
+	zoom_out_button.mouse_default_cursor_shape = (
+		Control.CURSOR_ARROW if zoom_out_button.disabled else Control.CURSOR_POINTING_HAND
 	)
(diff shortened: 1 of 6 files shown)


Kiisu-Master:
> **Note 2**: For some reason I always have problem when opening GodSVG project the first time right after cloning. But opening a second time seems to correct 🤔

This is a godot problem with importing assets and it happens with almost all projects (not a GodSVG specific thing).

thiagola92:
> This is a godot problem with importing assets and it happens with almost all projects (not a GodSVG specific thing).

Hmmm good to know, thanks

MewPurPur:
It's not like viewport.gd is a super complex file, but in this particular case it makes sense to make the zoom widget more reusable. Also the zooming code is a hot mess - so it's fine and I think this is the path going forward.

src/ui_parts/viewport.gd
```diff
@@ -71,32 +41,43 @@ func center_frame() -> void:
...
 		view.queue_redraw()
+	
 	if event is InputEventMouseMotion and\
 	event.button_mask in [MOUSE_BUTTON_MASK_LEFT, MOUSE_BUTTON_MASK_MIDDLE]:
 		set_view(view.position - event.relative)
 	
 	if event is InputEventPanGesture:
 		if event.ctrl_pressed:
-			zoom_level *= 1 + event.delta.y / 2
+			%ZoomMenu.zoom_level *= 1 + event.delta.y / 2
```
MewPurPur, line 51: Create an onready statically typed variable for ZoomMenu.

src/ui_parts/display/menu/zoom_menu/zoom_menu.gd
```diff
@@ -0,0 +1,54 @@
...
+
+
+const MIN_ZOOM = 0.125
+
+const MAX_ZOOM = 64.0
+
+signal zoom_changed(zoom_level: float)
+
+signal zoom_reseted()
+
```
MewPurPur, line 11: Add onready variables for the buttons in this file too.

Code style nitpicks:

- No whitespace between contextually related lines (like min zoom and max zoom, or the signals).
- One whitespace between contextually similar functions, two otherwise.
- zoom_in() and zoom_out() should be next to each other.

src/ui_parts/viewport.gd
```diff
@@ -71,32 +41,43 @@ func center_frame() -> void:
...
+
+func _on_zoom_changed(zoom_level: float) -> void:
+	var old_size_2d_override := size_2d_override
+	size_2d_override = size / zoom_level
+	set_view(view.position + (old_size_2d_override - size_2d_override) / 2.0)
+	display.material.set_shader_parameter(&"uv_scale",
+			nearest_po2(int(zoom_level * 32)) / 32.0)
+	update_view_limits()
+	controls.zoom = zoom_level
+	display_texture.zoom = zoom_level
```
MewPurPur, line 82: Can you check if changing the zoom variable in these two nodes can be centralized to the new zoom widget instead?

src/ui_parts/display/menu/zoom_menu/zoom_menu.gd
```diff
@@ -0,0 +1,54 @@
+extends HBoxContainer
+
+
+const MIN_ZOOM = 0.125
+
+const MAX_ZOOM = 64.0
+
+signal zoom_changed(zoom_level: float)
+
+signal zoom_reseted()
```
MewPurPur, line 10: Past time of "reset" is still "reset". Also, I see resetting doesn't emit the "zoom_changed" signal, which I find concerning but I haven't tested it.

MewPurPur (commented):
Code review, haven't tested the PR.

src/ui_parts/viewport.gd
thiagola92: The preference is `@onready` over `@export`?

MewPurPur: `%ZoomMenu` is a node path reference to a scene-unique node. If I rename ZoomMenu to ZoomWidget, for example, I'll have to change all these references.

src/ui_parts/viewport.gd
thiagola92: Like this?  

```gdscript
# handles_manager.gd

# ...

@onready var zoom_menu: HBoxContainer = %ZoomMenu

# ...

func _draw() -> void:
	# error because doesn't know the zoom_level type
	var thickness := 0.85 / zoom_menu.zoom_level
```

I thought about two solutions for this:  
1- We could give a type to the `ZoomMenu` instead of using `HBoxContainer`.  
2- We could have a "temporary/soft settings" to hold the zoom information instead of letting the widget/display have it.  

```gdscript
# temp_settings.gd
class_name TempSettings

signal zoom_changed(zoom: float)

var zoom: float = 1.0:
  set(value):
    zoom = clampf(value, MIN_ZOOM, MAX_ZOOM)
    zoom_changed(zoom)
```

(More tests would be needed in this case)

src/ui_parts/display/menu/zoom_menu/zoom_menu.gd
thiagola92: > Past time of "reset" is still "reset".  

Renaming to `zoom_reset_pressed` because `zoom_reset` is in use, but I'm open to better names.  

> Also, I see resetting doesn't emit the "zoom_changed" signal, which I find concerning but I haven't tested it.  

The method `zoom_reset()` change the `zoom_level` which emit `zoom_changed`.  

EDIT: I started a review? not my intention (don't know how to stop)

MewPurPur:
I think this is OK now. How do you want to be credited in the About section? "thiagola92" or "Your Real Name (thiagola92)"?

thiagola92:
> I think this is good for now. How do you want to be credited in the About section? "thiagola92" or "Your Real Name (thiagola92)"?

thiagola92 is good

MewPurPur:
Thanks!

PR #186, opened 2023-11-06
Declare zoom_menu variable at the start

thiagola92:
Last PR I removed `get_parent()` and inserted `%ZoomMenu`, but I believe that using `@onready` follow the project style.  

**Note**:  
Last PR I talked about giving a `class_name` to `ZoomMenu`/`ZoomWidget` scene, this could help with autocomplete.  

```gdscript
# Yes, awkward because type should be more than enough (probably a bug?)
@onready var zoom_menu: ZoomMenu = %ZoomMenu as ZoomMenu

# ...

func _draw() -> void:
	# will actually suggest zoom_level/zoom_in()/zoom_out()
	var size: Vector2 = get_parent().size / zoom_menu.zoom_level
```

### DIFF
--- a/src/ui_parts/view_camera.gd
+++ b/src/ui_parts/view_camera.gd
@@ -8,13 +8,16 @@ const ticks_interval = 4
 
 var surface := RenderingServer.canvas_item_create()  # Used for drawing the numbers.
 
+@onready var zoom_menu: HBoxContainer = %ZoomMenu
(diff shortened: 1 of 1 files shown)


MewPurPur:
No, don't add class_name to things like these. We'll use `const Weapon = preload("res://weapon.gd")`

PR #190, opened 2023-11-08
Remove unnecessary whitespace

Swarkin:
image


MewPurPur:
Rejecting due to: https://github.com/MewPurPur/GodSVG/blob/main/CONTRIBUTING.md#code-style

The reason for the code style guideline is that empty lines are harder to work with than ones that retain their scope - for example, you have to add the tabs back if you want to put a debug print statement.

PR #217, opened 2023-11-17
No need to rebuild TagSVG with every svg text change

SeremTitus:
Adds a diff tool that tracks text changes and maps actions from the change to TagSVG


MewPurPur:
I'll take a look at the code later. Could you summarize the infrastructural changes?

In general, my current mindset around text changes is that we shouldn't try to use diffs to try and inspect what the user changed, but we should instead treat every change to text that came from typing (including importing or pasting text inside) as something that should cause a full rebuild. Because changes to text can be unexpected.

So I don't think this will be useful for UndoRedo, I'd rather work with TextEdit's own UndoRedo when it comes to text changes, and use the usual UndoRedo outside of it.

Anyway, I'm not entirely opposed to at least having a tool to check for attribute diffs to avoid rebuilds, as those are relatively straightforward.

SeremTitus:
> I'll take a look at the code later. Could you summarize the infrastructural changes?
> 
> In general, my current mindset around text changes is that we shouldn't try to use diffs to try and inspect what the user changed, but we should instead treat every change to text that came from typing (including importing or pasting text inside) as something that should cause a full rebuild. Because changes to text can be unexpected.
> 
> So I don't think this will be useful for UndoRedo, I'd rather work with TextEdit's own UndoRedo when it comes to text changes, i.e. when the TextEdit is focused, and use the usual UndoRedo outside of it.
> 
> Anyway, I'm not opposed to at least having a tool to check for attribute diffs to avoid rebuilds, as those are relatively straightforward compared to changes of the SVG's overall tag layout.

The major change is moving  method  replace_self() in TagSVG to Tag, since in some cases I need to replace the tag,
the other add addition that compares string svg to get the changes and act on TagSVG  to reflect the new string svg... am sure after testing you may reconsider full rebuild stands

MewPurPur:
Please rebase this to a more recent version. I've encountered some regressions while testing but I can't tell which of them are caused by your PR as opposed to the dated version of GodSVG.

SeremTitus:
rebased

MewPurPur:
No code inspection yet, but from messing around, I'm having all sorts of unacceptable regressions. I've posted a video of me using GodSVG for a minute and encountering them. Notably, I got a load of errors like these:

> Tag.gd:26 @ set_unknown_attributes(): Signal 'propagate_value_changed' is already connected to given callable 'RefCounted(TagUnknown.gd)::emit_attribute_changed' in that object.
  <Stack Trace>  Tag.gd:26 @ set_unknown_attributes()
                 SVG.gd:110 @ update_using_svg_diff()
                 code_editor.gd:88 @ _on_code_edit_text_changed()"

and a crash at the end:

> Invalid get index 'd' (on base: 'Dictionary').

https://github.com/MewPurPur/GodSVG/assets/85438892/8216d933-5e81-46a1-9645-d9907b34d086

MewPurPur:
> prerequisite for features like undoredo

I've made it work by using only the SVG text: https://github.com/MewPurPur/GodSVG/pull/259

SeremTitus:
> > prerequisite for features like undoredo
> 
> I've made UndoRedo work by using only the SVG text: #259

okay, then this feature will not be necessary I will put turn it into draft and close it after a day or two, using svg text only to implement undoredo is a good approach it will not be able to tell which action exactly was done but it will work

PR #218, opened 2023-11-17
Implement move_tags_to in TagSVG

SeremTitus:
- prerequisite for features like drag and drop

### DIFF
--- a/src/data_classes/TagSVG.gd
+++ b/src/data_classes/TagSVG.gd
@@ -163,8 +163,19 @@ func move_tags_in_parent(tids: Array[PackedInt32Array], down: bool) -> void:
 	tags_moved_in_parent.emit(parent_tid, old_indices)
 	tag_layout_changed.emit()
 
-#func move_tags_to(tids: Array[PackedInt32Array], pos: PackedInt32Array) -> void:
-	#pass  # TODO implement this.
+func move_tags_to(tids: Array[PackedInt32Array], to: PackedInt32Array) -> void:
+	var new_parent_tids:PackedInt32Array = Utils.get_parent_tid(to)
+	var new_parent_tag:Tag = get_by_tid(new_parent_tids)
+	var count = 0
+	for pos in tids:
+		var parent_tids:PackedInt32Array = Utils.get_parent_tid(pos)
+		var parent_tag:Tag = get_by_tid(parent_tids)
+		var tag:Tag = parent_tag.child_tags[pos[-1]]
+		new_parent_tag.child_tags.insert(to[-1] + count,tag)
+		parent_tag.child_tags.pop_at(pos[-1])
+		count += 1
+	tags_moved_to.emit(to,tids)
+	tag_layout_changed.emit()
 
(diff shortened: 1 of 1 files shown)


SeremTitus:
> If you want to take a shot at implementing this, you should use the `tags_moved_to` signal and make it work with the possibility of selecting multiple tags (e.g. with Ctrl+Click). You'd also need to have the Indications signleton listen for this signal and update those selections. This is why I left it for later, it's a really difficult task not just from a UI standpoint. I wouldn't try to tackle it until I have the UI to check various tricky scenarios with tag nesting, selections at different depths and and such, to make sure they all work.

am not sure if I completely understand you but  I have switched to `tags_moved_to`

SeremTitus:
> If you want to take a shot at implementing this, you should use the `tags_moved_to` signal and make it work with the possibility of selecting multiple tags (e.g. with Ctrl+Click). You'd also need to have the Indications signleton listen for this signal and update those selections. This is why I left it for later, it's a really difficult task not just from a UI standpoint. I wouldn't try to tackle it until I have the UI to check various tricky scenarios with tag nesting, selections at different depths and and such, to make sure they all work.

ohh, allow multiple tags

MewPurPur:
> You'd also need to have the Indications singleton listen for this signal and update those selections. This is why I left it for later, it's a really difficult task not just from a UI standpoint. I wouldn't try to tackle it until I have the UI to check various tricky scenarios with tag nesting, selections at different depths and and such, to make sure they all work.

You must still update `Indications.gd` to connect to this signal and update the selections. I'm also almost certain this won't work, for one, you don't take into consideration that a tag you're moving might have child tags, which should also have their Tag IDs updated, also you're using `add_tag()` which should only be used when you actually add tags.

MewPurPur:
`<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><g><g><circle cx="4" cy="4" r="1"/></g><circle cx="7" cy="8" r="1"/></g><circle cx="2.5" cy="12.5" r="1"/><circle cx="10.5" cy="2.5" r="1"/></svg>`

Basic example of a SVG with nesting that can be used for testing a UI.

SeremTitus:
> > You'd also need to have the Indications singleton listen for this signal and update those selections. This is why I left it for later, it's a really difficult task not just from a UI standpoint. I wouldn't try to tackle it until I have the UI to check various tricky scenarios with tag nesting, selections at different depths and and such, to make sure they all work.
> 
> You must still update `Indications.gd` to connect to this signal and update the selections. I'm also almost certain this won't work, for one, you don't take into consideration that a tag you're moving might have child tags, which should also have their Tag IDs updated, also you're using `add_tag()` which should only be used when you actually add tags. You also need to take into consideration that if a tag and its child tag is selected, only the parent needs to move. Check out how this logic is implemented in the above functions.
> 
> this isn't really testable without some kind of UI that implements the ability to move tags

the tag child are considered since am pop the tag it self it will move with its children, am working on drag and drop currently and child tag move with there parent tag

SeremTitus:
- implemented in #224

PR #219, opened 2023-11-17
Rework session saving

### LINKED ISSUE #123: Viewbox is automatically coupled in situations it shouldn't be.
Viewbox is automatically coupled at the beginning, changing the viewbox and corrupting SVGs a lot of the time.

Viewbox also doesn't automatically decouple when you change the viewbox from the code editor.

MewPurPur:
Removes session settings; instead, everything that must be retained between sessions now goes in save_data directly.

Also fixes #123


SeremTitus:
@MewPurPur  hey, this got read off child_tags_container

MewPurPur:
@SeremTitus This container is now added via code only if the tag has children.

PR #240, opened 2023-11-20
Make `import_warning_dialog` more usable

Swarkin:
The import warning dialog cut off the error texts before with no way of scrolling to view them fully:
image

Now, after the edits:
image

I've wrapped the text label into a ScrollContainer to allow for two-way scrolling, if the extra space still isn't enough.


MewPurPur:
Thanks!

PR #264, opened 2023-11-28
Save last used directory and last path command relative setting

### LINKED ISSUE #263: Certain things aren't saved, but should be.
The path command popup should remember when you select relative.

The export/import dialogs should remember the location on your PC they were last at, and only use the default location as a backup.

The zoom editors and camera should NOT do that (the default setup is good enough).

Kiisu-Master:
closes #263


MewPurPur:
Thanks!

PR #267, opened 2023-11-29
Fix context menu positioning

Kiisu-Master:
- Always get mouse position with `get_global_mouse_position()` because it takes fractional scaling into consideration.
- Make menu appear to **the right of the mouse** like in other programs. (it would center on x axis before, the change is in Utils.gd)


MewPurPur:
Thanks!

PR #282, opened 2023-12-01
Mouse wrapping

### LINKED ISSUE #280: Blender style mouse wrapping when navigating viewport
Its hard to move long distances when zoomed in.
Mouse wrapping makes it easier as then you wont have to let go and move the mouse every time you reach the border of the viewport.

Kiisu-Master:
Closes #280 

Progress
- [x] wrap the mouse around viewport bounds
- [x] relative camera movement with wrapping
- [x] toggle wrapping option in settings
- [x] english translation
- [x] german translation
- [x] bulgarian translation
- [x] clean and readable code

This uses the same technique as [godots mouse wrapping](https://github.com/godotengine/godot/blob/d76c1d0e516fedc535a2e394ab780cac79203477/core/input/input.cpp#L813)  in the viewport.


Kiisu-Master:
I have an idea that I'll try tomorrow

Kiisu-Master:
I tried my best to copy godots c++ mouse wrapping code to gdscript.
I tried 2 different methods before this.
1. had correct relative motion but would have random jumps sometimes because of mouse position getting out of sync somehow.
2. would ignore the relative movement when wrapping to not have camera jumps, which would make the cursor drift away from where it originally was clicked in the viewport.

Godots code still has the drift but only when the mouse is moving really really fast when wrapping (it seems to have a customizable margin for detection, idk how it exactly works).
This method works better in the godot editor because it has faster updates and the relative motion is smaller.
I think the same accuracy can be achieved in GodSVG when disabling v-sync for unlimited framerate.

**Does this need a code comment mentioning that its copied from godot?**

translations/translation_sheet.csv
```diff
@@ -2,6 +2,8 @@
 #language,Language,Език,Sprache
 #invert_zoom,Invert zoom direction,Обърни посоката на увеличение,Zoomrichtung umkehren
 #invert_zoom_desc,Swaps zoom in and zoom out with the mouse wheel,Разменя увеличението и намалянето с колелцето на мишката,Vertauscht Hinein- und Herauszoomen mit dem Mausrad
+#wrap_mouse,Wrap mouse,,Maus einschließen
+#wrap_mouse_desc,Wraps mouse when moving in the viewport,,Schließt die Maus beim bewegen innerhalb des Fensters ein
```
MewPurPur, line 6: "Wraps the mouse cursor around when panning the viewport"

translations/translation_sheet.csv
```diff
@@ -2,6 +2,8 @@
 #language,Language,Език,Sprache
 #invert_zoom,Invert zoom direction,Обърни посоката на увеличение,Zoomrichtung umkehren
 #invert_zoom_desc,Swaps zoom in and zoom out with the mouse wheel,Разменя увеличението и намалянето с колелцето на мишката,Vertauscht Hinein- und Herauszoomen mit dem Mausrad
+#wrap_mouse,Wrap mouse,,Maus einschließen
+#wrap_mouse_desc,Wraps the mouse cursor around when panning the viewport,,Schließt die Maus beim bewegen innerhalb des Fensters ein
```
MewPurPur, line 6: Превъртане на курсора

Превърта курсора когато той достига краищата на екрана

src/ui_parts/viewport.gd
```diff
@@ -100,3 +100,24 @@ func adjust_view() -> void:
 func _on_size_changed() -> void:
 	if is_node_ready():
 		adjust_view()
+
+func wrap_mouse(relative: Vector2) -> Vector2:
+	var view_rect:Rect2 = get_visible_rect().grow(-1.0)
```
MewPurPur, line 105:
```suggestion
	var view_rect: Rect2 = get_visible_rect().grow(-1.0)
```

src/ui_parts/viewport.gd
```diff
@@ -100,3 +100,24 @@ func adjust_view() -> void:
...
+	var mouse_pos: Vector2 = get_mouse_position()
```
MewPurPur, line 106:
```suggestion
	var mouse_pos := get_mouse_position()
```

src/ui_parts/viewport.gd
```diff
@@ -100,3 +100,24 @@ func adjust_view() -> void:
...
+	
+	var warp_margin := view_rect.size * 0.5
+	
+	var relative_sign := Vector2i(
+		1 if relative.x >= 0.0 else -1, 
+		1 if relative.y >= 0.0 else -1)
```
MewPurPur, line 112: Are you aware of Vector2.sign()? Looks like a more appropriate thing to use here.

Kiisu-Master: I was not aware

MewPurPur:
Thanks!

PR #296, opened 2023-12-07
Improve file workflows

### LINKED ISSUE #290: Add Ctrl + S saving to active file
It would be nice to be able to save with ctrl + s, as this is natural for most people and would probably speed up workflows if people can just ctrl + s and then check how the file looks. The implementation would look like this

- When a user saves to a file or opens a file that filepath is stored somewhere
- When the user hits ctrl + s the stored file's data is overwritten with whatever changes the user has made

Kiisu-Master:
Closes #290
What I have added:
- Save **dialog** with ctrl + s.
  A good compromise between possible accidental saving and usability i think.
- Prefill last opened file name in save file dialog.
  File name is stored for session.
<hr/>

Other features to add in the future (not this pr):
- Recently opened files with ctrl + shift + o

### DIFF
--- a/src/SVG.gd
+++ b/src/SVG.gd
@@ -93,11 +93,9 @@ func native_file_import(has_selected: bool, files: PackedStringArray,
 _filter_idx: int) -> void:
 	if has_selected:
 		apply_svg_from_path(files[0])
-		GlobalSettings.modify_save_data(&"last_used_dir", files[0].get_base_dir())
 
 func non_native_file_import(file_path: String) -> void:
 	apply_svg_from_path(file_path)
-	GlobalSettings.modify_save_data(&"last_used_dir", file_path.get_base_dir())
 
 
 func _on_files_dropped(files: PackedStringArray):
@@ -108,6 +106,8 @@ func apply_svg_from_path(path: String) -> int:
 	var svg_file := FileAccess.open(path, FileAccess.READ)
 	var error := ""
 	var extension := path.get_extension()
+	
+	GlobalSettings.modify_save_data("last_used_dir", path.get_base_dir())
 	
 	if extension.is_empty():
 		error = "#file_open_empty_extension"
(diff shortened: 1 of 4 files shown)


MewPurPur:
> The 3 file workflows that need to be improved

They need to be usable, i.e. the added features shouldn't make things way worse for any of them.

src/ui_parts/export_dialog.gd
```diff
@@ -47,30 +47,33 @@ func native_file_export(has_selected: bool, files: PackedStringArray,
...
+	ExportDialog.open_save_dialog(extension, native_file_export, non_native_file_export)
+
+static func open_save_dialog(file_extension: String, native_save_callback: Callable, non_native_save_callback: Callable):
 	# Open it inside a native file dialog, or our custom one if it's not available.
 	if DisplayServer.has_feature(DisplayServer.FEATURE_NATIVE_DIALOG):
-		DisplayServer.file_dialog_show("Export a ." + extension + " file",
-				Utils.get_last_dir(), "." + extension, false, DisplayServer.FILE_DIALOG_MODE_SAVE_FILE,
-				["*." + extension], native_file_export)
+		DisplayServer.file_dialog_show("Export a ." + file_extension + " file",
+				Utils.get_last_dir(),GlobalSettings.current_file_name + "." + file_extension, false, DisplayServer.FILE_DIALOG_MODE_SAVE_FILE,
```
MewPurPur, line 61:
```suggestion
				Utils.get_last_dir(), GlobalSettings.current_file_name + "." + file_extension, false, DisplayServer.FILE_DIALOG_MODE_SAVE_FILE,
```

src/SVG.gd
```diff
@@ -109,6 +112,9 @@ func apply_svg_from_path(path: String) -> int:
 	var error := ""
 	var extension := path.get_extension()
 	
+	GlobalSettings.current_file_name = path.get_file().rstrip("." + path.get_extension())
```
MewPurPur, line 115: Use `remove_suffix()` instead.

MewPurPur:
Set it as ready for review when it's ready for testing.

src/SVG.gd
Kiisu-Master: ok but what's the difference

MewPurPur: Something like "my_svg.svg" will get stripped to "my_"

MewPurPur:
Thanks!

PR #300, opened 2023-12-08
Small improvements to file opening.

Kiisu-Master:
- Adds drag and drop for files and file extension checks.
- Applies file extension checks and error messages when opening from cmdline
- makes all menus and dialogs extend `Dialog` and adds a `is_open `static variable to make sure that dropped files dont get opened when there is a dialog open.
- Moves opening SVG files to the SVG script

### DIFF
--- a/src/SVG.gd
+++ b/src/SVG.gd
@@ -2,6 +2,9 @@
 ## The SVG text, and the native [TagSVG] representation.
 extends Node
 
+const AlertDialog := preload("res://src/ui_parts/alert_dialog.tscn")
+const ImportWarningDialog = preload("res://src/ui_parts/import_warning_dialog.tscn")
+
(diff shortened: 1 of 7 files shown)


MewPurPur:
Is this still WIP?

Kiisu-Master:
It's working but..
Error message translations are missing and there is some code to deduplicate in the error system. Right now it's different checks that have the exact same code to create the popup, with the only difference being the error message.
Maybe something like this would be better?:
```gdscript
var error: String = ""
if invalid_svg_condition:
    error = "error message"
if not error.is_empty():
    # create error popup with error as text here
```

MewPurPur:
It doesn't matter if translations are missing, obviously.

MewPurPur:
Thanks!

PR #303, opened 2023-12-08
Implement cli file opening

ilikefrogs101:
This PR will implement file opening from the command line with a few lines of code, it should also hopefully allow you to open svg files with many file managers "Open With" feature.

_Production edit: Closes #303_

### DIFF
--- a/src/SVG.gd
+++ b/src/SVG.gd
@@ -16,7 +16,20 @@ func _ready() -> void:
 	SVG.root_tag.child_attribute_changed.connect(update_text)
 	SVG.root_tag.tag_layout_changed.connect(update_text)
 	
-	if GlobalSettings.save_data.svg_text.is_empty():
+	var open_file := false
+	var cmdline_args := OS.get_cmdline_args()
+	if cmdline_args.size() > 0:
+		var svg_file := FileAccess.open(cmdline_args[0], FileAccess.READ)
+		if svg_file != null:
+			open_file = true
+	
+	if open_file:
+		var svg_file := FileAccess.open(cmdline_args[0], FileAccess.READ)
+		var svg_text := svg_file.get_as_text()
+		text = svg_text
+		saved_text = svg_text
+		update_tags();
+	elif GlobalSettings.save_data.svg_text.is_empty():
 		root_tag.attributes.width.set_value(16.0, Attribute.SyncMode.SILENT)
 		root_tag.attributes.height.set_value(16.0, Attribute.SyncMode.SILENT)
 		update_text(false)
@@ -24,14 +37,6 @@ func _ready() -> void:
 		text = GlobalSettings.save_data.svg_text
 		saved_text = GlobalSettings.save_data.svg_text
 		update_tags()
-	var cmdline_args := OS.get_cmdline_args()
-	if cmdline_args.size() > 0:
-		var svg_file := FileAccess.open(cmdline_args[0], FileAccess.READ)
-		if svg_file != null:
-			var svg_text := svg_file.get_as_text()
-			text = svg_text
-			saved_text = svg_text
-			update_tags();
 	UR.clear_history()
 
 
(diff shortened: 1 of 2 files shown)


MewPurPur:
What you explained in the issue sounds cool.

> Ive done some experiments 6 months or so ago and I figured out this also enables file managers to use your program as the default for opening X file type, in your case svg.

This means that in some operating systems, this will make GodSVG appear in lists of programs capable of handling SVGs, to make it possible to open them by default? If that's the case, that's really cool!

ilikefrogs101:
I am not entirely sure if it will show up in the list because Ive got no clue how that is handled but Ik it will allow you to set it as the program for opening svg if it will only through a custom file path link thingy

src/ui_parts/code_editor.gd
```diff
@@ -19,6 +19,14 @@ func _ready() -> void:
...
 	SVG.root_tag.tag_layout_changed.connect(auto_update_text)
 	SVG.root_tag.changed_unknown.connect(auto_update_text)
+	try_open_with()
+
+func try_open_with() -> void:
+	await get_tree().process_frame
+
+	var cmdline_args := OS.get_cmdline_args()
+	if(cmdline_args.size() > 0):
+		apply_svg_from_path(cmdline_args[0])
```
MewPurPur, line 29: Style nitpick: Remove the parentheses after `if` as per the GDScript style guide. Also, please add a tag whitespace on the empty line (or don't have an empty line, it's a small enough function that it's readable anyway).

This code should probably go in SVG.gd when the SVG text itself is added. And without the await.

MewPurPur:
If we get this through and merged, how do you want to be added to authors?

> Real Name (ilikefrogs101)

or just

> ilikefrogs101

?

src/ui_parts/code_editor.gd
ilikefrogs101: alright, Ill fix those. C# habbits sneaking in lol

ilikefrogs101:
Just ilikefrogs101 please

src/ui_parts/code_editor.gd
ilikefrogs101: Reason its not in svg.gd is because (chances are ive interpreted the code wrong) I think it needs to run on _ready/frame after ready

ilikefrogs101: I see what you mean about SVG.gd now, Ill do that

ilikefrogs101: Done

MewPurPur:
Move the logic before the previous logic that sets up the SVG text.

ilikefrogs101:
thats what I originally had but I felt it would lead to messy if statements

ilikefrogs101:
perhaps I can do it with a bool though

ilikefrogs101:
that might also get wierd actually

ilikefrogs101:
The reason I had it at the end is Im not particularly happy about 
`svg_file := FileAccess.open(cmdline_args[0], FileAccess.READ)` twice but I suppose I can just change to this for the second one
`var svg_text := FileAccess.open(cmdline_args[0], FileAccess.READ).get_as_text()`

src/SVG.gd
```diff
@@ -16,7 +16,19 @@ func _ready() -> void:
 	SVG.root_tag.child_attribute_changed.connect(update_text)
 	SVG.root_tag.tag_layout_changed.connect(update_text)
 	
-	if GlobalSettings.save_data.svg_text.is_empty():
+	var opening_file := false
```
MewPurPur, line 19: This variable seems unnecessary.

Also lower down, remove the semicolon.

ilikefrogs101: the opening_file variable? It was the best way I could think of lol, how would you have done that?

ilikefrogs101:
I am now deeply confused because things I havnt touched are causing errors

src/SVG.gd
MewPurPur: Ah sorry, I didn't see that this was important for the condition below! In that case, just fix the semicolon.

ilikefrogs101: you wouldnt happen to know a direct clone from the repo with no changes gives me errors would you lmao?

ilikefrogs101: anyway Ill start a new PR because the fork this one is ffrom imploded in on itself

PR #314, opened 2023-12-12
Implement autoformatter to minify attributes

### LINKED ISSUE #293: Implement the ability to minify (not optimize) attributes
After much consideration, I've decided it would be useful to separate the concepts of minifying and optimizing SVGs. Minifying would definitely not use a third party, while optimizing still might.

Minifying is a subset of optimizations that don't change the layout of the SVG, but make certain attributes more or less compact, i.e. `#0000ff` may become `#00f` or `blue` to save characters. The most important minifying is in pathdata.

Unlike optimizations, which would happen on a button press, minifying would happen when you change the attribute. However, when you're editing the SVG code, you wouldn't want the text to change behind you. The minifying would be part of the parsing, so the moment the focus exits the TextEdit, it will happen automatically. Importing might need to focus the TextEdit when it happens, so that the minifying isn't as abrupt.

Minifying would need its settings, which would all be off by default:

**Numeric optimizations:**
- Remove leading zeroes (0.25 -> .25)

**Color optimizations:**
- Optimize 6-number hex codes into 3-number (#ddeeff -> #def)
- Optimize hex codes into named colors when possible (#f00 -> red)

**Path optimizations:**
- Remove unnecessary spaces before numbers
- Remove the letters of identical consecutive commands (or line after move)
- Remove spacing after arc flags

This would get GodSVG most of the way to good SVG optimization, and would set that feature back to low priority.

### LINKED ISSUE #313: rgb(r, g, b) format not supported in color attributes
`rgb(255, 255, 255)` doesn't work if you input it inside a color editor. But it is a valid way to represent white.

MewPurPur:
Resolves #293

Fixes #313

image


src/ui_parts/settings_menu.tscn
```diff
@@ -20,102 +21,82 @@ corner_radius_top_right = 2
...
 tooltip_text = "#invert_zoom_desc"
-focus_mode = 0
-mouse_default_cursor_shape = 2
 text = "#invert_zoom"
+section_name = "input"
+setting_name = "invert_zoom"
 
-[node name="WrapMouse" type="CheckBox" parent="MainPanel/VBoxContainer/TabContainer/#input_tab/VBox"]
-unique_name_in_owner = true
+[node name="WrapMouse2" parent="VBoxContainer/TabContainer/#input_tab/VBoxContainer" instance=ExtResource("4_2qeh2")]
```
Kiisu-Master, line 74:
```suggestion
[node name="WrapMouse" parent="VBoxContainer/TabContainer/#input_tab/VBoxContainer" instance=ExtResource("4_2qeh2")]
```

PR #335, opened 2023-12-16
Update translation_sheet.csv

Swarkin:
Add missing German translations

`#remove_spacing_after_flags` still needs to be translated.


MewPurPur:
Thanks!

PR #343, opened 2023-12-20
Remove version support for web build

aladvs:
There is now a var at the beginning of the JS file allowing it to be re-enabled if need be. No need for version metadata now. 🙄


MewPurPur:
Thanks!

PR #348, opened 2023-12-22
[Draft] Add translation guide draft as if this project was converted to gettext

Vovkiv:
You can check it here: https://github.com/Vovkiv/convert-to-gettext/tree/main/translations


Vovkiv:
Seems that some section links doesn't work, but in overall that how it might look

MewPurPur:
The PR workflow section has a lot of overlap with the "Translations" section now. We could rewrite by having a "Setup" section before the "Translations workflow" one to explain about Godot, cloning, etc. The translations section itself definitely suffers from length and a lack of introduction other than the generic setup, as well as from being partial - e.g. we shouldn't use wording like "pro tip", "please", and absolutely not the likes of "it's simple", "simply", "as easy as" when writing documentation.

Vovkiv:
> The PR workflow section has a lot of overlap with the "Translations" section now. We could rewrite by having a "Setup" section before the "Translations workflow" one to explain about Godot, cloning, etc. things that apply to both coding and translating.

I think it might be more beneficial to do similar thing that was did (and I later expanded) here: https://github.com/mjakeman/extension-manager/tree/master/po

("po" folder in GTK software often used to store po and pot files).

There we added readme file where I explained how exactly build, add and test translation, and how later make pull request. And documentation file was referenced in main readme. 
So, I think it should be good idea to split code contributing from translations. Place readme inside "translations" folder. Github will automatically render readme there.

> The translations section itself definitely suffers from length

I don't think this should be issue. I would prefer that documentation would cover more then less, so there would be less situations when user don't know how to do something so they forced to google something or try to reach out developers/maintainers.

> a lack of introduction into the system other than the generic setup

Well, there still missing information on how to actually test translation in Godot, what version of Godot is used (so user wouldn't download version 4.3, when projects is in 4.2; Same should be applied to your current code contributing guidelines - they don't mention what version of Godot is used.), what files should be included for pull request (because Godot likes to do changes inside .import files when they opened on new device, sometimes rendering entire project or part of it broken, like missing dependencies.). There definitely room to improve.

Also, I don't think I follow on "generic setup"

> and from odd language choices - e.g. we shouldn't use wording like "pro tip", "please", and absolutely not the likes of "it's simple", "simply", "as easy as" when writing documentation.

Well, yes, I was writing this at late time, so there many wording like this. I think this should be addressed once entire documentation is overall ready, so wording can be addressed at same time with typos and stuff like that.

MewPurPur:
I've updated CONTRIBUTING.md with a setup section. I believe the way to go forward is to add a new file in the translations folder, and to simply link to it from CONTRIBUTING.md

Vovkiv:
> Could you do that?

Yes, I can (but probably later, since I don't have stable internet connection for now).

> I've updated CONTRIBUTING.md with a setup section. I believe the way to go forward is to add a new file in the translations folder called INSTRUCTIONS.md, and to simply link to it from CONTRIBUTING.md

File should be called readme.md instead because I doubt that github (or any other git service to that matter) will render it as part of repository unless user specifically open INSTRUCTIONS.md file manually, which is something I prefer to avoid, but overall yeah, that's way to go

Vovkiv:
Here, v2 of possible guide. I also included .gdignore to screenshots folder so Godot will not import them.
Reworded here and there, added how to test, how to make godot recognize new language.
Now it look more close to usable.

MewPurPur:
Could you link to parts of https://docs.godotengine.org/en/stable/tutorials/i18n/localization_using_gettext.html to outsource the Godot-specific parts of this documentation as much as possible?

MewPurPur:
> They also link example of using c++ with xgettext (which is even not that relevant for Godot usage of gettext since Godot can generate pot's by itself and this is actually preferred way).

Yeah, that part confused me. You can link to specific paragraphs though. For example the one for how to get gettext, or how to use it for syncing. But I guess it makes sense to not fragment things too much.

> What do you think about? This might benefit everyone who plans on starting with gettext.

What do you mean by wiki page? If you're referring to Godot's documentation page, I mean all power to you to propose that on the godot-docs repo, but Godot docs are user-facing, so I doubt they'd write something intended for people that aren't Godot users or for external tools.

translations/readme.md
```diff
@@ -0,0 +1,265 @@
+## Translations
+
+You can help by contributing to translation(s)! Don't worry, even if you totally new to software translations, this guide got you covered!
+
+It's recommended to use [Poedit](https://poedit.net/download/) to work on translation, but using nothing other then [text editor](#working-on-translation-without-additional-tools) is also valid.
```
MewPurPur, line 5: I believe the intro section should have be a general quick overview of the system.

```suggestion
For translations, GodSVG uses gettext. To edit translations, it's recommended to use [Poedit](https://poedit.net/download/), but using a simple text editor is also a viable option.

[Introduction into the translation system, like what .po and .pot files are]
```

translations/readme.md
```diff
@@ -0,0 +1,265 @@
...
+
+2. Go to toolbar, `Project Settings...`
+
+![](screenshots/13_toolbar.png)
+
+3. `Localization` -> `POT Generation` -> `Generate POT`
+
+![](screenshots/14_navigate_to_pot_generation.png)
+
+4. You will be prompted to save `pot` file. Overwrite `GodSVG.pot` with it.
```
MewPurPur, line 21: This section should be able to be outsourced to the Godot docs.

translations/readme.md
```diff
@@ -0,0 +1,265 @@
...
+
+1. Go to toolbar, `Project Settings...`
+
+![](screenshots/13_toolbar.png)
+
+2. `Localization` -> `Translations` -> `Add...`
+
+![](screenshots/15_navigate_to_po_adding.png)
+
+You will be prompted to provide `po` file of your translation.
```
MewPurPur, line 35: This section should also be outsourced to the Godot docs.

MewPurPur:
@Vovkiv What do you think of this PR now that the PR for converting to gettext was merged?

PR #365, opened 2024-01-05
Export Presets for Android and etc2_astc=true

Alex2782:
Can Android stay inside? 

`etc2_astc=true` is necessary for Android, I'm not sure what effect it has on other systems.

Bildschirmfoto 2024-01-05 um 20 16 29


MewPurPur:
How would you like to be included in the AUTHORS list? With or without your real name?

Alex2782:
Maybe later, when I find time to optimize the app for Android. Username would be enough.

Issue #160, opened 2023-11-02
Zoom (InputEventMagnifyGesture) on Android


Alex2782:
v4.2.beta2.official [f8818f85e]

Tested on an Android tablet, `InputEventMagnifyGesture` does not arrive there, is it an engine problem or? If so I could try to debug and fix under Android Studio
https://github.com/MewPurPur/GodSVG/blob/5a009656f999506e08077ab4ed2223dc40afe5da/src/ui_parts/viewport.gd#L79

MewPurPur:
No idea, I haven't tested this app on phone for a while now. It would be appreciated if you can help!

Alex2782:
It's an "engine" problem, I'm closing here.

Logcat , "ScaleGestureDetector" but no `InputEventMagnifyGesture`
```
2023-11-02 14:46:45.599 31971-31971 ScaleGestureDetector    org.godotengine.bugreport2           I  TwScaleGestureDetector
2023-11-02 14:46:45.611 31971-32109 godot                   org.godotengine.bugreport2           I  _unhandled_input:InputEventMouseMotion: button_mask=1 (Left Mouse Button), position=((224.9121, 210.453)), relative=((2.249121, -4.047473)), velocity=((45.53934, -260.7545)), pressure=0.00, tilt=((0, 0)), pen_inverted=(false)
2023-11-02 14:46:45.611 31971-32109 godot                   org.godotengine.bugreport2           I  _unhandled_input:InputEventScreenDrag: index=0, position=((224.9121, 210.453)), relative=((2.249121, -4.047473)), velocity=((45.53934, -260.7545)), pressure=0.00, tilt=((0, 0)), pen_inverted=(false)
2023-11-02 14:46:45.611 31971-32109 godot                   org.godotengine.bugreport2           I  _unhandled_input:InputEventScreenDrag: index=1, position=((188.4764, 441.1588)), relative=((-1.7993, 2.698297)), velocity=((-211.065, 139.1805)), pressure=0.00, tilt=((0, 0)), pen_inverted=(false)
2023-11-02 14:46:45.611 31971-32109 godot                   org.godotengine.bugreport2           I  4.6330361366272
2023-11-02 14:46:45.616 31971-31971 ScaleGestureDetector    org.godotengine.bugreport2           I  TwScaleGestureDetector
[...]

PR #366, opened 2024-01-05
enable_pan_and_scale_gestures

Alex2782:
https://github.com/MewPurPur/GodSVG/issues/160#issuecomment-1792865080

this setting is necessary on Android to be able to zoom with gesture control


MewPurPur:
Thanks!

PR #374, opened 2024-01-09
fix #345  Disable wrap mouse option if not supported

### LINKED ISSUE #345: Disable the "Warp mouse" setting, if it is not available.
`Control.warp_mouse()` is not always available, for example Web doesn't allow it, and Wayland doesn't support it yet AFAIK (I haven't tested this). So on these platforms, the setting should be disabled and maybe some kind of warning should be given.

Kiisu-Master:
This pr does not include separate error message for that yet, im too lazy.
I haven't tested this so i just hope the display server is correct.

hopefully fixes #345 

also added back the await in svg because i think its still needed? don't remember anymore but i think that was a issue again


MewPurPur:
I don't think we need universal logic for this, i.e. this shouldn't be an exported property. Let's have this logic in settings_menu.tscn

src/ui_parts/settings_menu.tscn
```diff
@@ -3,7 +3,7 @@
 [ext_resource type="Script" path="res://src/ui_parts/settings_menu.gd" id="1_1gf4m"]
 [ext_resource type="Texture2D" uid="uid://c528knojuxbw6" path="res://visual/icons/Languages.svg" id="2_ndyp7"]
 [ext_resource type="Script" path="res://src/ui_elements/BetterTabContainer.gd" id="3_vslgx"]
-[ext_resource type="PackedScene" path="res://src/ui_elements/setting_check_box.tscn" id="4_2qeh2"]
+[ext_resource type="PackedScene" uid="uid://chgcy63ldo7x5" path="res://src/ui_elements/setting_check_box.tscn" id="4_2qeh2"]
```
Kiisu-Master, line 6: idk if i should remove this

MewPurPur: If it happened without you prompting it, don't remove it

MewPurPur:
Thanks!

PR #380, opened 2024-01-11
Small file workflow fix/improvement.

Kiisu-Master:
Change window title to reflect open file.
Change some function names to make more sense.
Fix file name changes before import confirmed.
Fix import returning same svg text it just received.


MewPurPur:
Superseded by the above mentioned PR

PR #381, opened 2024-01-11
Add theme to HScrollBar

Kiisu-Master:
This copies the VScrollBar styleboxes to HScrollBar.
They are not linked though.
Only one that is different is the background because of the content margins.


MewPurPur:
Thanks!

MewPurPur:
Linking stuff will happen when we have custom theme generation logic.

PR #385, opened 2024-01-11
Add file to window title and improve viewport navigation.

Kiisu-Master:
- Add the currently open file path to the window title.
- Add ctrl + middle mouse zooming.
- Add comments to viewport input handling.
- Change viewport input handling to be else-if cases for event types.

### DIFF
--- a/src/ui_parts/viewport.gd
+++ b/src/ui_parts/viewport.gd
@@ -5,6 +5,9 @@ const ZoomMenuType = preload("res://src/ui_parts/zoom_menu.gd")
 const buffer_view_space = 0.8
 const zoom_reset_buffer = 0.875
 
+# Holds zoom position for ctrl + middle mouse zooming.
+var _zoom_to: Vector2
+
 @onready var display: TextureRect = %Checkerboard
 @onready var view: Camera2D = $ViewCamera
 @onready var controls: Control = %Checkerboard/Controls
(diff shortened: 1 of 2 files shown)


src/ui_parts/viewport.gd
```diff
@@ -55,22 +58,44 @@ func _unhandled_input(event: InputEvent) -> void:
...
+		else:
+			set_view(view.position - (wrap_mouse(event.relative)\
+					if GlobalSettings.wrap_mouse else event.relative) / Indications.zoom)
 	
-	if event is InputEventPanGesture:
+	elif event is InputEventPanGesture:
+		
+		# Zooming with ctrl + touch?
 		if event.ctrl_pressed:
 			zoom_menu.set_zoom(Indications.zoom * (1 + event.delta.y / 2))
```
Kiisu-Master, line 82: Idk what this is

MewPurPur:
Since the logic is under InputEventMouseMotion, can you do this with one of the properties for checking for relative movement instead of with a variable?

Kiisu-Master:
The variable is storing the position it will zoom to.
I don't see a property in the event that contains the starting position, so i dont think i could remove it.
This is not a relative thing. Zooming amount is using the relative movement, also with warping, but there is no way to know the position to zoom to in a stateless way.

MewPurPur:
Thanks!

PR #390, opened 2024-01-14
Remove all integer division warnings

DevPoodle:
This removes all integer division warnings by explicitly casting types, instead of having Godot implicitly cast them. It should make the code slightly more clear, and allows us to avoid the current issue with warning_ignore not working properly (see https://github.com/godotengine/godot/issues/56592).


MewPurPur:
Thanks!

PR #392, opened 2024-01-15
Make drag and drop autoscroll accelerate more near edges.

Kiisu-Master:
In case someone is confused, here is a graph.
m is the `scroll_area`.
https://www.desmos.com/calculator/msp5v3av4e

### DIFF
--- a/src/ui_parts/tag_container.gd
+++ b/src/ui_parts/tag_container.gd
@@ -1,7 +1,7 @@
 extends PanelContainer
 
 # Autoscroll area on drag and drop. As a factor from edge to center.
-const safe_margin := 1.0 / 2.0
+const scroll_area := 1.0 / 2.0
 const autoscroll_speed := 32.0
 
 @onready var scroll_container: ScrollContainer = $ScrollContainer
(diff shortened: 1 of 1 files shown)


Kiisu-Master:
Uhhh something is wrong with the safe margin.

MewPurPur:
So this doesn't work correctly in your PR yet?

Kiisu-Master:
It works fine. I just noticed that for some weird reason the zone in which autoscroll starts isn't of the correct size ( 1 / 6 ). It's a bit bigger, but also not twice as big. Idk why it is like that...

Kiisu-Master:
Also i believe this can still be improved. Right now its linearly increasing and max speed depends on the size of the scroll container.

Kiisu-Master:
give me a moment i have an idea

src/ui_parts/tag_container.gd
```diff
@@ -1,6 +1,8 @@
 extends PanelContainer
 
-const safe_margin = 1 / 6.0
+# Autoscroll area on drag and drop. As a factor from edge to center.
+const safe_margin := 1.0 / 2.0
```
Kiisu-Master, line 4: Maybe i should rename the variable... its not exactly the safe margin.
Its the scroll area.

MewPurPur: Sure.

Are you sure about 1/2? Seems quite extreme.

Kiisu-Master: 1/2 is actually 1/4 of the scroll container height.
Also keep in mind scrolling is slower near the center.

MewPurPur: I know, and I know, but it seems extreme nonetheless. You normally only have scrolling towards the very edges, the central part is for dropping things without repercussions.

MewPurPur:
Could you show a clip of how it looks?

src/ui_parts/tag_container.gd
MewPurPur: Tweaked it from 1/2 to 1/3, it seemed_ best to me that way. With 1/2, I had situations where I grab a tag and it immediately starts scrolling.

MewPurPur:
Tested it and I really like how it works!

Issue #292, opened 2023-12-06
Support for the transform attribute


MewPurPur:
This would require the following:

1. Add AttributeTransform (for attributes of type transform-list). This attribute would store a Transform2D.
2. Implement a TransformParser to parse between the string and the Transform2D implementation.
3. Implement an editor for this type of attribute. I'm not sure how this would work and look, so if you have any ideas, please comment on this post.
4. Add the attribute to all tags. This will be bloated, but it's necessary and other features later on will have to reduce this bloat in different ways.

PR #394, opened 2024-01-15
Add Support for TransformAttributes

DevPoodle:
Attempt at implementing https://github.com/MewPurPur/GodSVG/issues/292:
- [X] Add TransformAttribute
- [X] Add TransformParser
- [X] Add attribute to all tags, and add a basic editor for transforms
- [X] Update handles so they can transform along with the tag

_Production edit: Resolves #292_


MewPurPur:
Awesome!

The fifth checkbox isn't necessary for this PR, I think we're better off brainstorming that part separately instead of in the PR introducing transforms.

How do you want to be credited in authors? Just "DevPoodle" or do you also want to list your real name?

src/data_classes/TransformParser.gd
```diff
@@ -0,0 +1,21 @@
...
+static func transform_to_text(transform: Transform2D) -> String:
+	var output := "matrix(%s, %s, %s, %s, %s, %s)"%[transform[0].x, transform[0].y, transform[1].x, transform[1].y, transform[2].x, transform[2].y]
+	return output
+
+static func text_to_transform(text: String) -> Transform2D:
+	if text.is_empty():
+		return Transform2D.IDENTITY
+	var formatted_text := text.strip_edges()
+	formatted_text = formatted_text.trim_prefix("matrix(")
+	formatted_text = formatted_text.trim_suffix(")")
```
MewPurPur, line 12: An initial implementation of this function must account for all possible syntaxes of transform, including ones like "translate(2 0) rotate(45 8 8)"

MewPurPur: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform

DevPoodle:
I just go by DevPoodle. Also, thanks for the link, I think I've got all of the types of transforms implemented now.

MewPurPur:
Looks good, could you show how the transform field looks?

MewPurPur:
image

We can iterate on a better transform editor later in development, this will do for now. But not like this, it needs to be single-line like all other attribute editors. For now, you can just have a BetterLineEdit where you can type out the transform, with an arrow next to it to popup this matrix editor.

src/ui_parts/handles_manager.gd
```diff
@@ -636,21 +643,20 @@ var was_handle_moved := false
...
 	if event is InputEventMouseMotion:
 		should_deselect_all = false
 		var event_pos: Vector2 = event.position / Indications.zoom +\
 				get_node(^"../..").view.position
 		if dragged_handle != null:
 			# Move the handle that's being dragged.
-			var new_pos := SVG.root_tag.world_to_canvas(event_pos)
-			if snap_enabled:
-				new_pos = new_pos.snapped(snap_vector)
+			var new_pos := dragged_handle.transform.affine_inverse() * (event_pos if !snap_enabled else event_pos.snapped(snap_vector))
```
Kiisu-Master, line 659: Moving handles doesn't work with viewbox anymore.
I tried putting the new position calculation inside the old `SVG.root_tag.world_to_canvas( ... )`, which accounts for all except the O1 O2 values for some reason.

DevPoodle: Sorry, I'm not quite sure what you mean. I just tried using viewbox and it seems fine, but I'm not too familiar with viewbox so I might just be missing it. What's happening that's unexpected?

DevPoodle:
Here's what transform-field looks like now:


https://github.com/MewPurPur/GodSVG/assets/102482062/92de57b1-d320-4521-a631-26a7cf5d20f6

src/ui_parts/handles_manager.gd
MewPurPur: Try it with a decoupled viewbox, i.e. different width and height than the viewbox.

src/ui_elements/transform_field.gd
```diff
@@ -0,0 +1,68 @@
+## An editor to be tied to a transform attribute.
+extends HBoxContainer
+
+signal focused
+var attribute: AttributeTransform
+var attribute_name: String
+
+const MatrixPopup = preload("res://src/ui_elements/matrix_popup.tscn")
+
+@onready var line_edit := $LineEdit
```
MewPurPur, line 10: Type it as a BetterLineEdit, otherwise Godot will only know it's a Node.

MewPurPur:
I'm not sure if this is a problem, but a lot of the diff I see in Files Changed is from different PRs.

Kiisu-Master:
> I'm not sure if this is a problem, but a lot of the diff I see in Files Changed is from different PRs.

Github wouldn't show changes in diff if its already there in main.

src/ui_parts/handles_manager.gd
Kiisu-Master: https://github.com/MewPurPur/GodSVG/assets/142301391/09f18d65-ed39-41fe-93e2-7a0b3ea4db6a

Kiisu-Master:
Right now the transform attribute is always added in the SVG code, even when it's the default value. I think that's not needed.

DevPoodle:
Alright, I think I've gotten in all of the suggested improvements so far. The commit history was also looking weird for me, with some repeated commits and some commits from other pr's, so I've just rebased it. Hopefully it's better now.

MewPurPur:
I'll merge this in a bit. There are a few small issues with the implementation, but I think it's best to merge early and iterate after.

PR #414, opened 2024-01-21
Preload tag icons for performance

DevPoodle:
I added a const dictionary in Utils that maps the names of tags to their icons. This improves performance and makes the code a little cleaner. While I did add the dictionary to Utils, it could just as easily go in one of the other autoloads instead if that would be preferred.

### DIFF
--- a/src/ui_parts/tag_editor.gd
+++ b/src/ui_parts/tag_editor.gd
@@ -1,7 +1,5 @@
 extends VBoxContainer
 
-const unknown_icon = preload("res://visual/icons/tag/unknown.svg")
-
 const ContextPopup = preload("res://src/ui_elements/context_popup.tscn")
 const TagEditor = preload("tag_editor.tscn")
 const TransformField = preload("res://src/ui_elements/transform_field.tscn")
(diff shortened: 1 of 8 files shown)


MewPurPur:
Nice idea; did you microbenchmark that section of the code to see if this has a noticeable effect?

MewPurPur:
Tested it and it's something like a 0.1-0.3% performance boost to rebuilding. I'm not sure it's worth it, but I'm fine with merging it for now.

MewPurPur:
Wait, I think this would be better as a constant on the tags!

src/ui_parts/tag_editor.gd
```diff
@@ -39,8 +39,7 @@ func _ready() -> void:
 	RenderingServer.canvas_item_set_z_index(surface, 1)
 	title_label.text = tag.name
 	Utils.set_max_text_width(title_label, 180.0, 0.0)  # Handle TagUnknown gracefully.
-	title_icon.texture = unknown_icon if tag is TagUnknown\
-			else load("res://visual/icons/tag/" + tag.name + ".svg")
+	title_icon.texture = tag.icon
```
MewPurPur, line 42: Doesn't this mean unknown_icon no longer needs to be a variable in tag_editor.gd?

MewPurPur:
Thanks!

PR #419, opened 2024-01-21
Fix function typo

Kiisu-Master:
This typo caused path command points added with the hotkeys to first appear at a wrong location.


MewPurPur:
Thanks!

PR #423, opened 2024-01-22
Improve editing of transform attributes

### LINKED ISSUE #417: Enhance transform editing (comments with ideas are welcome)
So to elaborate on the situation, transform-list attributes are a list of the following 6 things:

- matrix(x1 y1 o1 x2 y2 o2)
- rotate(d ox oy) or rotate(d) if ox = 0 and oy = 0
- transate(x y) or translate(x) if y = 0
- scale(x y) or scale(x) if x = y
- skewX(d)
- skewY(d)

This list of elements gets applied one after another to what's initially an identity matrix.

Currently, this is all provided as the final matrix, but matrices are sucky to work with. This post is for brainstorming how a good transform editor would look (not a visual editor in the viewport, that would be its own discussion).

MewPurPur:
Resolves #417


DevPoodle:
It looks like the compute_transform functions for skewX and skewY aren't quite right. For svg, skew uses the tangent of the input in degrees: (https://www.w3.org/TR/css-transforms-1/#SkewXDefined).

MewPurPur:
Thanks for the heads up. This PR is still WIP.

MewPurPur:
image

image

image

image

image

PR #462, opened 2024-01-31
Add context actions in the viewport

### LINKED ISSUE #323: Right Mouse Button on a handle should bring up the tag context popup
When you click RMB on handles, have the same context popup show up that would show up if you clicked the corresponding command or tag editor. It would be pretty helpful to your productivity.

### LINKED ISSUE #338: Right Mouse Button in the empty area of the viewport should bring up a context menu for adding a shape there.
This would be massively helpful for productivity.

### LINKED ISSUE #355: Context popup of tag multi-selection has Move Up/Down options erroneously
Ctrl+Click multiple tags and right click one of them. You'll see Move Up/Down appear as if your selection was only the clicked tag. Instead, Move Up/Down should only be available if the minimum subset of tags selected is at the same depth, and if they aren't the first/last tags in their parent tag.

MewPurPur:
Moves selection-dependent actions in Indications.gd

Simplifies the building blocks of the inspector, like path command editors, in doing so.

Resolves #323 
Resolves #338 
Fixes #355


MewPurPur:
Merging this despite a bug with popup positions, since all of the basic logic is in place.

PR #475, opened 2024-02-07
Convert translations to po files and use english for keys, add readme for editing translations.

### LINKED ISSUE #347: Replace current CSV translation system with gettext
This is kind of big step, but manageable one.
gettext currently most widely used system for translating software and games.
It has many nice features that every translator will appreciate! But might be less obvious then something like csv file.

What should be done:
1. Convert CSV to gettext. This is relatively easy task, most of time you need to replace keys like "tooltip_1" with your source text (so, effectively move text from csv file to your source scripts). 
2. Convert other translations (bulgarian, german(?), and others) to gettext. This is also relatively easy task, some copy pasting required.
3. Check if this thing even works, catch bugs if they arise.
4. Somehow teach/introduce you to gettext basics, that might be little harder, but also manageable.
5. Add translation documentation. It's not necessary has something to do with gettext, but this is still something that should be taken seriously if you want people to spend their time translating your software.

Kiisu-Master:
closes #347 and supersedes #348 
This converts the csv translation sheet to po and pot files, uses english for the translation keys and adds a readme file to explain the translation editing process.
Also adds a utility script `update_po_files.sh` to automate updating all po files after changing the pot.


MewPurPur:
Cool! A PR like this should also explain the new workflow, and I'd also like to hear about the rationale of each change.

Kiisu-Master:
There weren't many changes.
I created basic po files based on the csv we have.
And then i took all translation keys in the source code and replaced them with the english translation, because thats how you use gettext apparently.

Kiisu-Master:
I decided not to use Godots built in POT generator because it hasn't been properly updated in at least 3 years and has a lot of problems.

MewPurPur:
Could you give your PR a more descriptive name?

MewPurPur:
Something looks wrong with the long strings right now, am I mistaken?

translations/bg.po
```diff
@@ -0,0 +1,332 @@
...
+msgid "Paste"
+msgstr "Постави"
+
+msgid "Cut"
+msgstr "Изрежи"
+
+msgid "Alert!"
+msgstr "Предупреждение!"
+
+msgid ""
```
MewPurPur, line 237: Like this, is it normal?

Vovkiv: Believe it or not, but yes, tho, it's probably better to write it as:
```gettext
msgid "The file couldn't be opened.\n"
"Try checking the file path, ensure that the file is not deleted, or choose a "
"different file."
msgstr "Файлът не може да бъде отворен.\n"
"Опитайте да проверите пътя към файла, уверете се, че файла не е изтрит, или "
"изберете различен файл."
```
So it won't look that confusing

Kiisu-Master: It was formatted like that after i opened it in Poedit

Kiisu-Master:
Wouldn't using English for translation keys make it harder to change it?
You'd have to go trough the code base and replace every occurrence of it...
I know there arent many reused keys, but still it doesn't seem good.

Kiisu-Master:
I can't find `#syntax_error` in the translation sheet, so ill just add it to the POT.
https://github.com/MewPurPur/GodSVG/blob/9291f125e03198c7ab2ac634bdfeef2f5093aa7a/src/ui_parts/import_warning_dialog.gd#L42

Kiisu-Master:
Ok so technically it never even reaches this code. It crashes when the svg has wrong syntax because text_to_svg returns a stringname when its trying to create the preview image.

MewPurPur:
> I dont know what the point of creating an error message there even is, if you already crash and early return before that part.

The crash is only if you import a SVG, in other situations the error text works correctly.

I'll address this issue in a few minutes.

MewPurPur:
Needs a rebase after #540

MewPurPur:
Enough waiting for Godot. Is this ready for merging?

Kiisu-Master:
I didn't find any issues after merging right now so yeah

MewPurPur:
Thanks! :tada:

PR #485, opened 2024-02-11
Expand the functionality of drag-adding handles

### LINKED ISSUE #395: Improve working with bezier control handles
Cubic beziers, shorthand cubic beziers, and quadratic beziers are annoying to work with because of how the control handles overlap with the beginning or the end. We should implement a system where adding these commands places the control points in the middle, or in the one third and two thirds points of the line between the points for full cubic beziers.

MewPurPur:
**Half finished:** Make it so when you're doing things from the new viewport context popup, newly added path commands or tags get snapped to your mouse until you click, only then they get conclusively added.
**Half finished:** Make it so when you're adding curves in the viewport, the bezier handles don't overlap with the main handles. (Resolves #395)

The current idea is that when something is added, its handle gets flagged as being dragged. This no longer would apply to only path command handles, but also to XY handles like the ones in tags like \<circle>.

There's a catch though, that only that handle would be dragged, even though multiple handles might need to move, for example in \<line> and in the new curve logic. My current idea for solving this is to remove and re-add the tag every frame to recalculate how it would be added.

I'm also thinking that tags that can be added in the viewport might need some inner logic for "moving them" that takes a Vector2. This would resolve the difficulty with \<line> having two handles.

### DIFF
--- a/src/Indications.gd
+++ b/src/Indications.gd
@@ -22,7 +22,7 @@ signal selection_changed
 signal proposed_drop_changed
 
 # The viewport listens for this signal to put you in handle-placing mode.
-signal added_path_handle
+signal added_object
 
 # The PackedInt32Array holds the hierarchical orders. TID means Tag ID.
 # For example, the 5th child of the 2nd child of the root tag would be (1, 4).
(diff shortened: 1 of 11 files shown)


MewPurPur:
Nothing interesting here anymore.

PR #503, opened 2024-02-16
Update German translation

Kiisu-Master:


MewPurPur:
Is it ready to be merged?

Kiisu-Master:
Yes...
The spacing after flags translation is maybe not the best, but better than none i hope.
I added "(flag)" so it should be clear.

MewPurPur:
Thanks!

PR #504, opened 2024-02-16
Improve panning button mask

Kiisu-Master:
It would stop panning if both middle and left mouse buttons were held at once.
Both are valid panning buttons so i dont see a reason for it to be exclusive.

### DIFF
--- a/src/ui_parts/viewport.gd
+++ b/src/ui_parts/viewport.gd
@@ -57,7 +57,7 @@ func _unhandled_input(event: InputEvent) -> void:
 		return
 	
 	if event is InputEventMouseMotion and\
-	event.button_mask in [MOUSE_BUTTON_MASK_LEFT, MOUSE_BUTTON_MASK_MIDDLE]:
+	event.button_mask & (MOUSE_BUTTON_MASK_LEFT | MOUSE_BUTTON_MASK_MIDDLE):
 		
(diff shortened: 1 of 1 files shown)


src/ui_parts/viewport.gd
```diff
@@ -57,7 +57,7 @@ func _unhandled_input(event: InputEvent) -> void:
 		return
 	
 	if event is InputEventMouseMotion and\
-	event.button_mask in [MOUSE_BUTTON_MASK_LEFT, MOUSE_BUTTON_MASK_MIDDLE]:
+	event.button_mask & (MOUSE_BUTTON_MASK_LEFT | MOUSE_BUTTON_MASK_MIDDLE):
```
MewPurPur, line 60: `event.button_mask == MOUSE_BUTTON_MASK_LEFT or event.button_mask == MOUSE_BUTTON_MASK_MIDDLE`

Would prefer to not have bitwise stuff in the code if not necessary.

Kiisu-Master: Its all mask values... why not use bitwise operations for it?

MewPurPur: Oh I see, this won't work the way I suggested.

MewPurPur:
Thanks!

PR #515, opened 2024-02-19
Add file saving and loading for web platform.

### LINKED ISSUE #511: File dialog is broken in the web build
Let's say I'm using the web build, and I want to import or export something.
I get this:
image
At first glance, this seems fine, but it's actually not; because this is not any part of my PC.
I faintly recall the Godot docs saying something about web builds accessing a database with a filesystem inside; maybe that's what it's looking through?

Either way, I think the problem is that it opens Godot's file manager, when we really want the OS file manager.

Kiisu-Master:
Closes #511 

Todo list:
- [ ] Maybe make JS upload function general use. Just thinking about the reference image feature that was requested..
- [ ] Discuss where to put download code.
- [ ] Research if it's possible to fix it not working sometimes. Would be nice to find a way to make it more synchronous - without weird signal and waiting stuff. There was a callback used in the original extension, but it was removed in pixelorama. I also cannot test using JavaScript interfaces that would allow callbacks because 4.3 dev3 has a problem with that.
- [x] Add credits again.
- [x] Discuss if it should even be an autoload.
- [x] Handle file name correctly (this just a reminder, i forgot to add ` .get_file()` for svg download)


Pacomatic1:
>  Discuss if it should even be an autoload.

I don't think it should be an autoload, because for people not using the web build it'd be a waste of resources.
As such there are two things I can think of:

- Add a little check in svg.gd (or any other pre-existing autoload) that loads the script if you're on the web build,
- Or load the script when you import or export something presuming you're on web.
The main downside to the latter is making file I/O have a long delay and it's more prone to bugs if we do anything involving the file manager and forget to add it or something.
With that said, anyone using the web version would likely be importing and exporting anyways, so it probably won't be much an issue to make it automatically load on web.

MewPurPur:
HandlerGUI would be a good place for this.

Kiisu-Master:
So i copy the code to HandlerGUI and remove the separate file?

MewPurPur:
IMO yeah, that's the way to go.

MewPurPur:
I like the code itself.

MewPurPur:
Thanks!

PR #539, opened 2024-02-24
Add missing translation for "Syntax error"

Kiisu-Master:
It's used in the import warning dialog.
The code has never been triggered yet because there's both a crash and a early return before it, but if that get's fixed then this has to be added too.


MewPurPur:
Superseded by #540

Issue #496, opened 2024-02-14
Solidify the implementation of adding new objects


MewPurPur:
Concerns #395
Concerns #483
Concerns #484

Adding new objects would be one of two things:

- Adding new tag paths after RMB in the viewport
- Inserting new path commands after RMB in the viewport or using the shortcuts

We need some special logic for adding tags or path commands at a position, the regular logic won't do:
- Adding a new line at (x,y) should make it so the second handle is at (x + 1, y)
- Adding a new curve path command at should generate the control points between the start and the end, similar to how Convert To currently behaves.

Moreover, this special logic needs to run every frame. Seemingly, we'll need to implement a `temporary_add()` kind of function in a lot of places, one that doesn't concern UndoRedo.

When it comes to cancelling these changes, I believe what we actually need is to solidify some logic for cancelling actions, which needs its own proposal.

MewPurPur:
Would be really good to get this right before the next alpha.

PR #553, opened 2024-02-28
Improve logic of adding handles

MewPurPur:
Aims to solve #496 and #395


MewPurPur:
No, this ain't it... but there are nice things in here that will surely land in the final solution.

Closing this for now, I'll make a PR porting the best parts.

PR #558, opened 2024-02-29
Fix current file not updating when saving with ctrl + s

Kiisu-Master:
This now also includes small improvement that adds the "autoformatting options" as tooltip for the button.


MewPurPur:
Thanks!

PR #564, opened 2024-03-02
Fix the cyclic preload (corrupted scene) issue with a get function.

Kiisu-Master:
This should fix the occasional corrupted scene error and make this project runnable in Godot 4.3 .


MewPurPur:
Thanks!

PR #565, opened 2024-03-02
Make custom preload solution a static variable to prevent loading same scene multiple times.

Kiisu-Master:


MewPurPur:
Hmm, sure.

MewPurPur:
Thanks!

PR #570, opened 2024-03-06
Refactor savefiles and implement configurable shortcuts

### LINKED ISSUE #327: Allow to copy/paste palettes as text.
A TextEdit would make it faster to edit palettes. It could look something like this:

```
NamedColor1 = "fac"
NamedColor2 = "cfa"
NamedColor3 = "abcdef"
NamedColor4 = "rgb(42, 69, 123)"
```

with all the goodies like syntax highlighting when syntax isn't valid, indicators for how to do this, etc. Although the goodies can be added later on.

### LINKED ISSUE #542: Allow more customizability in the Settings dialog
This PR will need to be detailed in many small proposals and done very gradually.

What I have on my mind is to make the dialog a bit bigger and have the following tabs in it:

- Shortcuts tab, which exposes some shortcuts for customizing, like Ctrl+D and such. It also exposes shortcuts that can't be modified but are used in GodSVG for documentation, as well as to show you what's wrong by throwing errors like "[Shortcut] is already used by [unmodifiable action]". Unmodified actions would be things like the path command shortcuts.
- ~~Autoformatting settings get moved to the main settings in another tab. Instead of tabs, the different attribute types and the way to optimize them will be in a few sections. There's a text telling you that changing these settings immediately affects the current graphic.~~
- ~~Theming tab, which would be used for customizing various things. It would have at least two sections, one for the general theming colors and variables, one for the SVG highlighter colors.~~
- ~~Keep the palettes tab as is.~~
- ~~Add a last tab for the rest of the settings, like the settings in the current input tab.~~

### LINKED ISSUE #569: Exporting png with scaling just extends the png
I am using the latest released version of GodSVG for windows as of posting this issue.

I am encountering issues with exporting scaled vector drawings to png : Scaling just extends the png, it does not scale the actual drawing.

I do not think this is intended because this,
image

outputs this,
rocket

MewPurPur:
Shortcuts tab is now populated! (but in this draft it's not configurable yet)

Reworks the palettes system to be more straightforward. Now the palettes are in the same config file as the other settings, including the new shortcut configs. This means users' old palettes might poof after this PR is merged.

Fixes color wheel in the color picker not being draggable in RGB mode if the initial color is pure black or not displayable.

Fixes #569

Closes #542

Edit: Now also resolves #327


MewPurPur:
Okay, so I implemented #327 to make it easier for users to port their palettes. You can now paste XML in the palette settings to generate them automatically. It looks like so:

```xml
<palette title="Pure">
	<color value="#fff" name="White"/>
	<color value="#000" name="Black"/>
	<color value="#f00" name="Red"/>
	<color value="#0f0" name="Green"/>
	<color value="#00f" name="Blue"/>
	<color value="#ff0" name="Yellow"/>
	<color value="#f0f" name="Magenta"/>
	<color value="#0ff" name="Cyan"/>
</palette>
```

If you can find the palettes.tres file. It's going to have a few bigger lines which contain the title of your palette, while the colors themselves are going to be in a format that looks like so:

```
[sub_resource type="Resource" id="Resource_m1yax"]
script = ExtResource("1")
name = "Common icon"
color = "e0e0e0"

[sub_resource type="Resource" id="Resource_uwqg6"]
script = ExtResource("1")
name = ""
color = "808080"

[sub_resource type="Resource" id="Resource_u1q1i"]
script = ExtResource("1")
name = "Animation"
color = "c38ef1"

[sub_resource type="Resource" id="Resource_mtwjg"]
script = ExtResource("1")
name = "2D"
color = "8da5f3"

[sub_resource type="Resource" id="Resource_hl25n"]
script = ExtResource("1")
name = "3D"
color = "fc7f7f"

[sub_resource type="Resource" id="Resource_4h4vg"]
script = ExtResource("1")
name = "GUI Control"
color = "8eef97"

[sub_resource type="Resource" id="Resource_a0a0k"]
script = ExtResource("1")
name = "Mesh"
color = "ffca5f"
```

You can only keep the name and color, and fill them in here:

```xml
<palette title="Godot colors">
<color value="COLOR HERE" name="NAME HERE">
</palette>
```

If the name value is empty, don't fill it in. Then copy the second row as many times as necessary. So for our example, we'd get:

```xml
<palette title="Godot colors">
<color value="e0e0e0" name = "Common icon"/>
<color value="808080"/>
<color value="c38ef1" name = "Animation"/>
<color value="8da5f3" name="2D"/>
<color value="fc7f7f" name = "3D"/>
<color value="8eef97" name = "GUI Control"/>
<color value="ffca5f" name = "Mesh"/>
</palette>
```

This is still manual, but Find & Replace should make it easier for bigger palettes. If you want, you can just copy the values and names manually.

Issue #342, opened 2023-12-19
Dealing with GodSVG's blurriness


MewPurPur:
The current situation with GodSVG's scaling is very unhappy.

### Disabled scaling:

- It just makes the UI too small in my opinion. It's better for something to be too big, than to be forced too small.

### Canvas items scaling mode (current):

- It scales most things regularly, so they become blurry.

-----

The current idea is to move to disabled scaling mode, but I've decided that its downside is worse than that of canvas item scaling. So not until a setting for editor scale is added.

I imagine such a setting could go from something like `1x` to `2x`. I am not sure how it would be implemented: I imagine with lots of bubblegum and shoestrings. But the main things are:

- The theme needs to be updated whenever this setting changes. So for example for 2x, things like border width and corner radius need to be doubled (some extra detail might be needed for high corner radius too).
- A lot of things throughout the UI need to be updated too. Where do I even begin... Basically anything transform-related, for example minimum sizes, needs to be multiplied or tweaked in other ways manually, via script.
- The SVGs used for icons will likely need some logic around reimporting them to the correct size. (Note: Keep track of https://github.com/godotengine/godot/pull/86022)
- HandlerGUI could have a static variable "edscale" used for all the above.

I feel like some things may be doable, to some extent, in more sweeping ways than via script, but I don't know which ones, how, and if there would be any complications in using those ways.

colleen05:
I have a 1440p monitor at a higher PPI than your typical 21"-27" 1080p display. This means that things typically look smaller on my monitor if they were made for 1080p displays. In my opinion, even with the probably-smaller-than-intended UI elements, the UI elements in the default window size are perfectly fine!

I would recommend disabling scaling by default and having UI scaling options in the preferences. This would allow for better rendering code to be written in the future for cleaner large-scale UI (e.g., a 24x24 button at 2x scale is actually rendered at 48x48, rather than the window texture being scaled up with texture filtering).

I propose three scaling options:
1. **No scaling:** Everything is rendered to a texture the size of the window. The user may decide to user their OS's native window scaling.
2. **Factor scale:** The user selects a factor (e.g, 150%), and the UI is rendered to a texture that is equal to the window's size when scaled by the selected factor. I.e., a 1280x720 window with 200% factored scaling will have UI rendered at 640x360.
[...]

PR #571, opened 2024-03-09
Implement ThemeGenerator

MewPurPur:
Will be needed to allow the theme to be changed by the user cleanly (i.e. light theme, dark theme). Will also be needed for #342.

- Implements ThemeGenerator, just an object that runs quickly when the project is opened (takes like 15ms for me) and sets up the theme. It can be rerun again to change the theme, though for now there's no way to make it create a theme different than the current one. But it can be modified fairly easily.
- Improves the theme a little
- Small optimization to the shortcut settings page.


MewPurPur:
Now there's all the infrastructure for good theme generation. The rest of the work will happen in follow-ups.

PR #587, opened 2024-03-14
[WIP] Fix high GPU usage with custom camera implementation

### LINKED ISSUE #537: High GPU usage on idle
I noticed it when running the project with forward+ instead of compatibility, but i think it's doing the same on compatibility.

I think Godot is capable of preventing continuously redrawing, as the editor itself has a spinner to debug the editor redrawing, and there is a canvas item redraw debug visualisation available for debugging projects . When turning it on, it doesn't show anything updating on idle so i have no clue what's causing the GPU usage to go to 80%  :/

Kiisu-Master:
The grid does not work.
fix #537


MewPurPur:
Thanks!

PR #590, opened 2024-03-15
Fix toggling visuals after grid drawing fix.

Kiisu-Master:
The node reference was not updated in the display script after the code was moved back to camera.
Also had to decouple drawing and updating view.
Spaghetti code!!


MewPurPur:
You've not seen true spaghetti

MewPurPur:
Thanks!

PR #591, opened 2024-03-15
Update translation files

Kiisu-Master:
@Swarkin i added german translations please check
I dont like how everything has `farbe`, but its like that in english too


MewPurPur:
Thanks!

Swarkin (changes requested):
I left some comments on what could be changed

PR #614, opened 2024-03-22
Implement Clear Button

Qainguin:
This clear button is in the CodeEditor scene.

When pressed, it returns the SVG to the boilerplate.

If there is a better way of returning the SVG to boilerplate please let me know.

clear


visual/icons/Clear.svg
```diff
@@ -0,0 +1 @@
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="m8 1a1 1 0 0 0 -1 1v5h-2c-1.108 0-2 .892-2 2v1h10v-1c0-1.108-.892-2-2-2h-2v-5a1 1 0 0 0 -1-1zm-5 10v4l10-1v-3z" fill="#def"/></svg>
```
MewPurPur, line 1: `<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="m7 7h-2a2 2 0 0 0 -2 2v1h10v-1a2 2 0 0 0 -2-2h-2v-5a1 1 0 0 0 -2 0zm-4 4v4l10-1v-3z" fill="#def"/></svg>`

MewPurPur:
Seems like the monospace font was re-imported for some reason and its fallback disappeared. Is this change intentional?

Anyway, the approach looks good to me. I'd like to think about this a little more, because I'm not sure about having a destructive action be readily available with a click, and also the amount of buttons next to each other is becoming pretty big.

MewPurPur:
I think we should take a different approach: Make it so that Copy All Text and Reset SVG are two options of a ContextPopup that a single button opens. The context menu may later house more options, but I think that's better even if it's just two.

Qainguin:
> Seems like the monospace font was re-imported for some reason and its fallback disappeared. Is this change intentional?

This was not intentional and just happened when I opened the project

> I think we should take a different approach: Make it so that Copy All Text and Reset SVG are two options of a ContextPopup that a single button opens. More options might be in that menu later, but I think that's better even if it's just two.

What should the icon be for that button and what would the tooltip be?

MewPurPur:
I don't know, maybe no need for a tooltip for now. It could use More.svg, or something custom if you have any ideas. The context popup buttons could use the Copy svg and the new one you added

Qainguin:
I think this commit does what you wanted, I used a custom SVG because I felt that if I used More.svg it would look too similar to the one on the right.

MewPurPur:
You still need to revert the changes to FontMono.ttf.import

Qainguin:
That commit should have fixed it. I also deleted the CopyButton because I forgot to in the other commit.

MewPurPur:
And I think I'd prefer a different icon. I agree with your reasoning, but "</>" is closely associated with HTML. Maybe something like a pencil would work better? I'm not sure.

Qainguin:
Alright, I changed the icon to a pencil rotated 45 degrees.

visual/icons/CodeOptions.svg
```diff
@@ -0,0 +1 @@
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path transform="translate(7.5-3) rotate(45)" d="m7 1c-.554 0-1 .446-1 1v2h4v-2c0-.554-.446-1-1-1zm-1 4v7l2 3 2-3v-7zm1 1h1v5h-1z" fill="#def"/></svg>
```
MewPurPur, line 1:
```suggestion
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path transform="rotate(45 8 8)" d="M7 1a1 1 0 0 0 -1 1v2h4v-2a1 1 0 0 0 -1-1zm-1 4v8l2 3 2-3v-8zm1 1h1v6h-1z" fill="#def"/><path d="M9 6l-5 5" fill="none" stroke="#def" stroke-opacity=".4" stroke-width="2"/></svg>
```

Optimized and a bit tweaked to not be a carbon copy of Godot's Edit.svg

src/ui_parts/code_editor.gd
```diff
@@ -124,9 +127,22 @@ func _on_file_button_pressed() -> void:
 	Utils.popup_under_rect_center(context_popup, file_button.get_global_rect(),
 			get_viewport())
 
+
+func _on_options_button_pressed():
+	var btn_array: Array[Button] = [Utils.create_btn("Copy All Text", _on_copy_button_pressed, false, CopyIcon), Utils.create_btn("Reset SVG", reset_svg, false, ClearIcon)]
```
MewPurPur, line 132: Surround the text with tr() to make it translatable.

Qainguin:
Thanks, I updated the translatability and merged your CodeOptions.svg file.

visual/icons/CodeOptions.svg
Qainguin: I agree this does look way better, thanks!

MewPurPur:
Ok, I think it's ready now.

Qainguin:
I have marked it as ready.

MewPurPur:
Thanks!

Qainguin:
Final question, do I have to update the web build or will you do that?

src/ui_parts/code_editor.gd
```diff
@@ -124,9 +127,22 @@ func _on_file_button_pressed() -> void:
...
+	Utils.popup_under_rect_center(context_popup, options_button.get_global_rect(), get_viewport())
+
+
 func clear_file_path() -> void:
 	GlobalSettings.modify_save_data("current_file_path", "")
 
+func reset_svg() -> void:
+	SVG.text = '<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg"></svg>'
+	SVG.update_tags()
+
```
Qainguin, line 145: SVG.update_text(true)

Issue #639, opened 2024-04-06
Automated builds with github actions


Kiisu-Master:
Benefits i can think of:
- Easy testing of dev builds for people who can't build it themselves.
- Removes the need for a exported web build inside the repository.
- Reproducible builds are more trustworthy.
- Simplifies setting up of builds for new people, as they would have the workflow files as a reference.
- Instant feedback for when builds fail (not that important but still a benefit).

Implementation plan: look at how Pixelorama did it.
We could also add static checks but that seems a bit more complicated as currently we haven't enforced gdformat formatting or anything.

PR #640, opened 2024-04-06
Create automatic builds for Windows and Linux

Kiisu-Master:
Starting work for #639.
This adds automatic builds for Windows and Linux.
It downloads Godot from the godot builds repo.
WINE and rcedit are needed to add icons to the Windows executable.

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GODOT_VERSION: 4.2.1
+  GODOT_VERSION: 4.2.2.rc2
   EXPORT_NAME: GodSVG
 
 jobs:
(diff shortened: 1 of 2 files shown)


MewPurPur:
Can you explain what the PR's code does?

Kiisu-Master:
> Can you explain what the PR's code does? 

I think the step names explain enough... if you don't understand specific lines then ask

> By the way, the project is now being developed on 4.2.2.rc2

I haven't tried using it yet, I'm not sure if the docker thing supports Godot dev releases

MewPurPur:
Well in all honest I don't know what most lines mean, like if I need to update this file I wouldn't be able to, aside from maybe replacing all "4.2.1"s with the newest Godot version. Which by the way brings me to the question, could you explain what kinds of maintenance or improvements will this file need?

MewPurPur:
Thanks!

PR #643, opened 2024-04-09
Changes in preparation for global menu support

aaronfranke:
This PR makes a few changes and additions to the way actions are done in GodSVG. Now basically all functionality found in the top buttons also have an input action associated with them. Many of these input actions do not have a key bound to them, but they could be bound in the future, either out of the box, or allowing users to bind them.

Aside from input actions, this PR also adds a few signals, adds a few public methods to change state normally only accessible through buttons, and makes a few style enhancements.

These changes are designed to allow support for global menu integration in the future. However, I decided to future-proof and use the new API available only in Godot 4.3 or later, thus why I am splitting this into 2 PRs. For this first PR, everything in here is compatible with Godot 4.2.x. For what the full global menu support will look like, check out this branch: https://github.com/aaronfranke/GodSVG/tree/global-menu

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -15,7 +15,7 @@ config/version="v1.0-alpha2"
 config/tags=PackedStringArray("project")
 run/main_scene="res://src/ui_parts/main_scene.tscn"
 config/use_custom_user_dir=true
-config/features=PackedStringArray("4.2", "Forward Plus")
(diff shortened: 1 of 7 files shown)


MewPurPur:
Thanks!

PR #644, opened 2024-04-09
Add native global menu support (macOS)

aaronfranke:
Draft because I used the new Godot 4.3+ API so this will only work with Godot 4.3+. Follow-up to #643.

Screenshot 2024-04-09 at 4 33 47 AM
Screenshot 2024-04-09 at 4 34 07 AM
Screenshot 2024-04-09 at 4 34 17 AM
Screenshot 2024-04-09 at 4 34 29 AM

This is not just about clicking buttons in the menu, it also provides convenient documentation for the user of what actions are available and which have keys available to press.
Screenshot 2024-04-09 at 4 33 14 AM

### DIFF
--- a/src/ui_parts/global_menu.gd
+++ b/src/ui_parts/global_menu.gd
@@ -1,0 +1,223 @@
+extends Node
+
+
+var global_rid: RID
+var appl_rid: RID
+var help_rid: RID
+
+var file_rid: RID
+var file_index: int
+var file_clear_svg_index: int
+var file_optimize_index: int
+var file_clear_assoc_index: int
+var file_reset_svg_index: int
+
+var edit_rid: RID
+var edit_index: int
+var tool_rid: RID
+var tool_index: int
+
+var view_rid: RID
+var view_index: int
+var view_show_grid_index: int
+var view_show_handles_index: int
+var view_rasterized_svg_index: int
+
+var snap_rid: RID
+var snap_index: int
+var snap_enable_index: int
+var snap_0125_index: int
+var snap_025_index: int
+var snap_05_index: int
+var snap_1_index: int
+var snap_2_index: int
+var snap_4_index: int
+
+
+func _enter_tree() -> void:
+	if not DisplayServer.has_feature(DisplayServer.FEATURE_GLOBAL_MENU):
+		queue_free()
+		return
+	# Included menus.
+	global_rid = NativeMenu.get_system_menu(NativeMenu.MAIN_MENU_ID)
+	appl_rid = NativeMenu.get_system_menu(NativeMenu.APPLICATION_MENU_ID)
+	help_rid = NativeMenu.get_system_menu(NativeMenu.HELP_MENU_ID)
+	# Custom menus.
+	_generate_main_menus()
+	_setup_menu_items()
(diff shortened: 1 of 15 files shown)


MewPurPur:
Took a peek today, this seems too hard-coded. The shortcuts are only assigned on tree enter, and they don't seem to account for the user has changing the shortcuts. If necessary, you can add a custom notification for changed shortcuts in Utils.gd and listen to it here, fetching the first shortcut from each action instead.

I don't know the new 4.3 functionality though, maybe that's not possible.

MewPurPur:
Needs to be updated after #711.

Please check out the TranslationUtils class!

MewPurPur:
We're on a 4.3 beta now, this can be considered

aaronfranke:
> (wait, what happened to this PR?)

It looks like you somehow managed to edit a commit from way back in the history, and force-push, which modified all commits beyond that point too. Don't worry, I can fix it easily, that is, whenever I get around to working on this.

MewPurPur:
Okay, thanks! Yeah, I was doing some weird things to delete a file from the git history (a file which is no longer in the repo or even relevant, but made 80% of its size).

MewPurPur (changes requested):
TranslationUtils.get_shortcut_description() should be used where possible. Also, there's a SHORTCUTS_CHANGED custom notification now - the signal isn't needed. _action_call(action) could also be replaced with ShortcutUtils.fn(action).call().

Could you also replace the remaining tr() calls with TranslationServer.translate()? We decided recently to always use this version because it can be used in static functions.

Also open the "update_translations.gd" script to update translations automatically after you're done.

btw we're already on the 4.3 betas, so this will work.

aaronfranke:
> TranslationUtils.get_shortcut_description() should be used when it can.

Done.

> Also, there's a SHORTCUTS_CHANGED custom notification now - the signal isn't needed.

I'm not sure, it doesn't seem the same. Also, isn't a notification worse than a signal? A notification is propagated to everything, while a signal is only propagated to what connects to it.

> _action_call(action) could also be replaced with ShortcutUtils.fn_call(action).

I don't think so? `ShortcutUtils.fn_call` doesn't handle everything.

> Could you also replace the remaining tr() calls with TranslationServer.translate() instead?

Done.

> Also open the "update_translations.gd" script to update translations automatically after you're done.

Done (and I discarded changes to the po/pot files made before running this script, so the changes to the po/pot files in this PR are purely from the script).

src/ui_parts/global_menu.gd
```diff
@@ -0,0 +1,217 @@
...
+	NativeMenu.clear(help_rid)
+	NativeMenu.clear(file_rid)
+	NativeMenu.clear(edit_rid)
+	NativeMenu.clear(tool_rid)
+	NativeMenu.clear(view_rid)
+	NativeMenu.clear(snap_rid)
+
+
+func _setup_menu_items() -> void:
+	# Included App and Help menus.
```
MewPurPur, line 97: TranslationUtils should be used here to get the translated string of "open_settings". Same for the lines of code below. That way it's ensured that they are the same everywhere.

MewPurPur:
Thanks!

PR #649, opened 2024-04-10
Update automation

Kiisu-Master:
- Fix rcedit and wine not working
- Add comments
- Cache Godot binaries
- Run exports in parallel
- Report error if no artifacts are uploaded
- General overhaul

Didn't add the github pages automatic deploy yet because it probably needs a different workflow - it has permission requirements stuff that i don't understand yet. I also didn't want to risk accidentally running it from this pr, who knows what would happen.


MewPurPur:
Thank you!

PR #661, opened 2024-04-16
Implement UI scaling setting

Kiisu-Master:
Implemented UI scaling setting.
Also tried to fix the tooltips of other settings in the category. It was getting set on the wrong node.


MewPurPur:
Thanks!

PR #666, opened 2024-04-18
Add an "Auto UI scale" feature

aaronfranke:
As of #115, the default UI scale is unusable on macOS:
Screenshot 2024-04-17 at 7 25 07 PM

This PR adds a new feature, enabled by default, that will automatically increase the UI scale. I added a new checkbox to the settings menu. When unchecked, the UI scale number directly sets the UI scale. When checked, the UI scale number acts as a multiplier, and the actual scale is calculated from this number and the window size.
Screenshot 2024-04-17 at 7 26 49 PM


MewPurPur:
Thanks!

PR #671, opened 2024-04-19
Add new strings to GodSVG.pot in advance of global menu PR

aaronfranke:
I also snuck in the new signal in GlobalSettings.gd.


MewPurPur:
I think we need to do "Move to (Relative) and "Move to (Absolute)" to spare ourselves from 20 translations.

MewPurPur:
I don't think this pull request is relevant after #711. I'll still keep it as a reminder for the keybinds_changed signal (which is irrelevant as all shortcuts display only in things like ContextPopup and probably tooltips in the future, both of which will are generated on the spot).

MewPurPur:
Closing as superseded by #799.

PR #672, opened 2024-04-19
Switch auto UI scale to use screen DPI

Kiisu-Master:
- This makes auto scaling use screen DPI.
- Limits scale to a calculated max scale to avoid glitches when min size would be bigger than usable screen size.

This has to be tested on more devices. @aaronfranke
I only tested this on my phone and laptop, which both have low resolution screens. Different DPI though so it's tested a little bit and seems to work.

Previously it was making the default experience on phone even worse, because it made UI even smaller than it already was without automatic scale. I know phones arent really supported anyways, but I hope this fix also makes it better on desktop.

Edit: I did some testing on my laptop (1366x768,  DPI 96). Previous implementation would make UI too small on it. New one seems to work well. Also tested with the AMD virtual resulution to set it to higher resolutions (FHD, DPI 120), and it still works well.

### DIFF
--- a/src/GlobalSettings.gd
+++ b/src/GlobalSettings.gd
@@ -246,8 +246,9 @@ func get_validity_color(error_condition: bool, warning_condition := false) -> Co
 
 
 func update_ui_scale() -> void:
-	await get_tree().process_frame
 	var window := get_window()
+	if not window.is_node_ready():
+		await window.ready
 	
 	var usable_screen_size := Vector2(
 		DisplayServer.screen_get_usable_rect(DisplayServer.window_get_current_screen()).size
--- a/src/GlobalSettings.gd
+++ b/src/GlobalSettings.gd
@@ -293,7 +293,7 @@ func _calculate_auto_scale() -> float:
 			return 1.0
 		elif dpi <= 120:
 			return 1.25
-		elif dpi <=160:
+		elif dpi <= 160:
 			return 1.5
 		elif dpi <= 200:
 			return 2.0
--- a/src/GlobalSettings.gd
+++ b/src/GlobalSettings.gd
@@ -193,9 +193,7 @@ func _enter_tree() -> void:
 	load_settings()
 	load_user_data()
 	DisplayServer.window_set_mode(save_data.window_mode)
-	get_window().wrap_controls = true  # Prevents the main window from getting too small.
 	update_ui_scale()
-	get_window().size_changed.connect(update_ui_scale)
 	ThemeGenerator.generate_theme()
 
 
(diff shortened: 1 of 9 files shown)


src/GlobalSettings.gd
```diff
@@ -249,24 +247,59 @@ func get_validity_color(error_condition: bool, warning_condition := false) -> Co
...
+			return 0.75
+		elif dpi <= 96:
+			return 1.0
+		elif dpi <=160:
+			return 1.25
+		elif dpi <= 240:
+			return 1.5
+		elif dpi <= 480:
+			return 1.75
+		elif dpi > 480:
```
aaronfranke, line 299: Make it clear to the reader that at least one of these blocks will be executed when `dpi != 72`

```suggestion
		else:  # dpi > 480:
```

src/GlobalSettings.gd
```diff
@@ -249,24 +247,59 @@ func get_validity_color(error_condition: bool, warning_condition := false) -> Co
...
+		if dpi < 72:
+			return 0.75
+		elif dpi <= 96:
+			return 1.0
+		elif dpi <=160:
+			return 1.25
+		elif dpi <= 240:
+			return 1.5
+		elif dpi <= 480:
+			return 1.75
```
aaronfranke, line 298: These values are just a bit too low. `DisplayServer.screen_get_dpi(screen)` returns `163` on my system with 200% scaling, so I would expect `2.0` to be returned, but this code is returning `1.5`.

src/GlobalSettings.gd
```diff
@@ -249,24 +247,59 @@ func get_validity_color(error_condition: bool, warning_condition := false) -> Co
...
+	
+	var usable_screen_size := Vector2(
+		DisplayServer.screen_get_usable_rect(DisplayServer.window_get_current_screen()).size
+		# Subtract window decoration size.
+		 - (window.get_size_with_decorations() - window.size)
+		)
+	
+	# How much can window content size be multiplied by before it extends over the usable screen size.
+	var diff :=  usable_screen_size / window.get_contents_minimum_size()
+	var max_scale := floorf(minf(diff.x, diff.y)*4.0)/4.0
```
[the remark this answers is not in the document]
Kiisu-Master: It don't look like that floors the value (we dont want to round it up)

src/GlobalSettings.gd
Kiisu-Master: I adjusted the values to be roughly dpi/96. That's the DPI of my laptop screen and 1.0 looks fine on it, so i used it as a base.

src/GlobalSettings.gd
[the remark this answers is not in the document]
MewPurPur: Oh, right, that'd change the logic. Then could you just put spaces around the math operators, as per GDScript's style guide?

src/GlobalSettings.gd
aaronfranke: Different perspectives I guess, I am using a high DPI desktop display.

src/GlobalSettings.gd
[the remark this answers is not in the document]
Kiisu-Master: I investigated the function. It does have floor in it but it adds 0.5 which makes it round.
```cpp
double Math::snapped(double p_value, double p_step) {
	if (p_step != 0) {
		p_value = Math::floor(p_value / p_step + 0.5) * p_step;
	}
	return p_value;
}
```

src/GlobalSettings.gd
Kiisu-Master: Your DPI will return 2.0 now. 

I found that changing the scaling in Windows also changes the reported DPI. With that, it looks like scaling in GodSVG now roughly matches the OS scaling so it should be good now.

src/GlobalSettings.gd
```diff
@@ -247,26 +246,67 @@ func get_validity_color(error_condition: bool, warning_condition := false) -> Co
...
+	
+	var dpi :=  DisplayServer.screen_get_dpi(screen)
+	if dpi != 72:
+		if dpi < 72:
+			return 0.75
+		elif dpi <= 96:
+			return 1.0
+		elif dpi <= 120:
+			return 1.25
+		elif dpi <=160:
```
aaronfranke, line 296:
```suggestion
		elif dpi <= 160:
```

aaronfranke (approved):

MewPurPur:
Thanks A TON!

PR #677, opened 2024-04-20
Arguable improvements to GoodFileDialog

MewPurPur:
Adds some right-click actions. Uses drives instead of system dirs for the panel on the left. Small improvements in other parts of the codebase.

I'm done with this widget now, it's more than good enough given GodSVG's current quality.


src/ui_parts/good_file_dialog.gd
```diff
@@ -138,23 +147,40 @@ func set_dir(dir: String) -> void:
...
+	var second_section: Array[String] = []
+	drives_list.clear()
+	for drive_idx in DirAccess.get_drive_count():
+		var drive_path := DirAccess.get_drive_name(drive_idx)
+		if is_system_dir(drive_path):
+			first_section.append(drive_path)
+		else:
+			second_section.append(drive_path)
+	var drive_paths := first_section + second_section
+	print(drive_paths)
```
Swarkin, line 163: There's a debug print 🙂

Swarkin:
I personally think system dirs should be listes alongside drive names, just like in blender.

Swarkin:
The `set_dir` function fails if the path exists but cannot be opened.

MewPurPur:
> The set_dir function fails if the path exists but cannot be opened.

Any way to detect that?

Swarkin:
Yes, DirAccess.open() can return null, which is also exactly the error that happened to me which I sent on the Discord earlier.

PR #703, opened 2024-05-04
Fix export dialog (again)

Qainguin:
Removed redundant lines and print statements.


MewPurPur:
Hmm, something seems wrong here.

Qainguin:
> Hmm, something seems wrong here.

Yeah, I need to delete and fork again, give me 5 minutes.

PR #704, opened 2024-05-04
Fixed QualityHBox showing when it shouldn't have

Qainguin:
This also fixes it printing when it shouldn't have.


MewPurPur:
Thanks!

PR #718, opened 2024-05-05
Add file references to po translation files.

Kiisu-Master:
- Added file path and line number references to all automatically collected strings.
- Made message a class which should make it easier to extend.
- Now also runs `msgmerge `on pot file itself so it gets formatted the same way.
- @Vovkiv 👋


MewPurPur:
LGTM, the "../" part of all these paths can be removed.

MewPurPur:
Thanks!

Vovkiv:
🥲

PR #719, opened 2024-05-06
Implemented "View In List"

WeaverSong:
Adds a "View In List" option when right-clicking a handle. This will scroll the tag list so that the selected tag is at the top of the view. If the selected handle is an inner command, the scroll will be offset based on its position, such that the first inner command will have the selected tag at the top of the view.


MewPurPur:
I suggest making it so if you RMB on a handle, it passes some context to the method inside Indications.gd, for example "context". The new option should only be shown when you've pressed a handle, which can be done with this argument.

> If the selected handle is an inner command, the scroll will be offset based on its position, such that the first inner command will have the selected tag at the top of the view.

I'd say, we should have a fixed offset for where the selected tag would be, for example 20% of the screen. The metric you've mentioned here can be user-defined, for example with a lot of unrecognized tags, to the point where the path command can end up outside of the screen.

src/Indications.gd
```diff
@@ -481,7 +492,10 @@ func get_selection_context(popup_method: Callable) -> ContextPopup:
 					can_move_up = true
 				if not can_move_down and base_tid[-1] < parent_child_count - filtered_count:
 					can_move_down = true
-		
+		if context == SELECTION_CONTEXT.VIEWPORT:
+			btn_arr.append(Utils.create_btn(TranslationServer.translate("View In List"),
+					view_in_list.bind(selected_tids[0]), false, load("res://visual/icons/ViewInList.svg")))
```
MewPurPur, line 497: There's a bit of oddness in doing this with `selected_tids[0]` when you have a multiselection, but it's not that important. I'll squash this eventually.

src/Indications.gd
```diff
@@ -456,8 +463,12 @@ func insert_inner_after_selection(new_command: String) -> void:
 			path_attrib.insert_command(last_selection + 1, new_command)
 			normal_select(semi_selected_tid, last_selection + 1)
 
+enum SELECTION_CONTEXT {
```
MewPurPur, line 466: The name of the enum must be PascalCase.

MewPurPur:
You need to run the update_translations.gd script to add the new string.

MewPurPur:
Thanks!

Issue #715, opened 2024-05-05
Hitting tab after editing a single property does not cause the property to update


WeaverSong:
If you edit a single property and then hit tab, the property is not updated and unless you remember to manually go back, re-select the textbox and hit enter, the changes you made there will be lost the next time something is updated.

MewPurPur:
Is this fixed?

MewPurPur:
Seems fixed, I'm closing it.

Issue #710, opened 2024-05-05
Bad ordering of tag attributes


MewPurPur:
An earlier PR messed up the ordering of attributes. We should implement TagContent widgets for each tag that simply sort the default attributes correctly, until we figure out how to make more complex widgets.

PR #720, opened 2024-05-07
Implements navigation using tab and shift+tab

WeaverSong:
Allows using tab and shift+tab to move forwards and backwards through standard and unknown tag attributes.

Also makes tab count as submitting text, closing #715 

Note that this PR does not touch Path commands; that's a whole complex system that I do not fully understand yet, and it's not clear to me how to go about implementing this in there.

### DIFF
--- a/src/ui_elements/BetterLineEdit.gd
+++ b/src/ui_elements/BetterLineEdit.gd
@@ -21,7 +21,7 @@ func _ready() -> void:
 	focus_entered.connect(_on_focus_entered)
 	focus_exited.connect(_on_focus_exited)
 	mouse_exited.connect(_on_mouse_exited)
-	text_submitted.connect(_on_text_submitted.unbind(1))
+	text_submitted.connect(_on_text_submitted)
 
 func _input(event: InputEvent) -> void:
 	if has_focus():
--- a/src/ui_elements/BetterLineEdit.gd
+++ b/src/ui_elements/BetterLineEdit.gd
@@ -21,7 +21,7 @@ func _ready() -> void:
 	focus_entered.connect(_on_focus_entered)
 	focus_exited.connect(_on_focus_exited)
 	mouse_exited.connect(_on_mouse_exited)
-	text_submitted.connect(func(): if not Input.is_action_just_pressed("ui_focus_next"): release_focus.unbind(1))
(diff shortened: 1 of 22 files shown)


src/ui_elements/BetterLineEdit.gd
```diff
@@ -21,7 +21,7 @@ func _ready() -> void:
 	focus_entered.connect(_on_focus_entered)
 	focus_exited.connect(_on_focus_exited)
 	mouse_exited.connect(_on_mouse_exited)
-	text_submitted.connect(release_focus.unbind(1))
+	text_submitted.connect(func(): if not Input.is_action_just_pressed("ui_focus_next"): release_focus.unbind(1))
```
MewPurPur, line 24: I'd prefer to do this in its own method, instead of a lambda. Also, this should probably include the ui_focus_previous action too?

MewPurPur:
I'd prefer to isolate the changes to BetterLineEdit in their own PR, as I can approve that more easily. I'm not convinced about needing an extra variable that requires setup in each attribute editor. I'm thinking the tag editor should by itself be able to set up the next and the previous focus.

WeaverSong:
I'd gone with the extra variable to avoid making the tag editor access child nodes of the attributes, but it does make sense for it to be handling all of the focus stuff.

MewPurPur:
> to avoid making the tag editor access child nodes of the attributes

Oh, I see now. Then yeah, we can go with your approach. I really want to somehow turn all these attribute editors into single nodes if possible, it'd be a considerable optimization and reduction of complexity

WeaverSong:
After some more thought, I believe it might be possible to do with without the extra variable in each attribute, by instead making tab frames, tag contents and attributes "focusable", and then passing off their focus to the appropriate child, as well as passing on the focus_previous and focus_next to the appropriate children. I'll do the split you suggested, and then see if I can get a more elegant solution working with the focus system.

MewPurPur:
I was planning to make attributes individually focusable at some point, but it'd be after alpha 5.

Maybe I'll resolve this together with #710

WeaverSong:
That makes sense to me. Either way it looks like this PR is outdated and to be implemented in a better way later on, so I'll close it now.

PR #721, opened 2024-05-07
Implemented tab submits

WeaverSong:
Makes tab and shift+tab in BetterLineEdits count as submitting the text.

### DIFF
--- a/src/ui_elements/BetterLineEdit.gd
+++ b/src/ui_elements/BetterLineEdit.gd
@@ -18,9 +18,9 @@ func _init() -> void:
 	caret_blink_interval = 0.6
 
 func _ready() -> void:
-	focus_entered.connect(_on_focus_entered)
-	focus_exited.connect(_on_focus_exited)
-	mouse_exited.connect(_on_mouse_exited)
+	focus_entered.connect(_on_base_class_focus_entered)
+	focus_exited.connect(_on_base_class_focus_exited)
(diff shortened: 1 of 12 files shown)


src/ui_elements/BetterLineEdit.gd
```diff
@@ -57,6 +59,10 @@ func _on_focus_exited() -> void:
 		text = text_before_focus
 		text_change_canceled.emit()
 
+func _on_text_submitted(_submitted_text) -> void:
+	if not Input.is_action_just_pressed("ui_focus_next") && not Input.is_action_just_pressed("ui_focus_prev"):
```
MewPurPur, line 63: As per GDScript's style guide, use "and" instead of &&

MewPurPur:
This seems dangerous. The _on_text_submitted function is found in a lot of inheriting classes. They could override this and prevent it from being called. Check

MewPurPur:
Maybe it would be better to generally avoid having "meaningfully named" methods for text_submitted responses and similar, to make it so this problem is avoided generally. Maybe change the method name to `_on_base_class_text_submitted`? Would super appreciate if you do this to other methods too, if you find any in BetterTextEdit and BetterLineEdit.

MewPurPur:
Looks good to me, could you just fix the whitespace noise in number_field.gd?

MewPurPur:
Added the style changes myself.

Thanks!

PR #723, opened 2024-05-09
Update Ukrainian and Russian translations

Vovkiv:
Also 2 more newlines in readme. I guess for compatability reasons better to create newlines after text but before new header?
Any markdown experts here?


MewPurPur:
Thanks!

PR #725, opened 2024-05-09
Allow moving view while dragging handle

### LINKED ISSUE #713: Allow moving view with middlemouse while dragging a handle.
Just a minor thing, but currently you cannot move the view while dragging a handle; so if the place you want to put the handle is off the screen, you have to drop it, move the view, then drag it again.

Kiisu-Master:
Fixes #713 

This feels like a hack to me but it seems to work and its a really small change IDK what to think of it


MewPurPur:
Thanks!

PR #730, opened 2024-05-11
Fix attributes getting wrongly added

### LINKED ISSUE #726: Crash when adding line tag
Crash when adding line tag.
Also crashes on empty tag in file:
```svg
<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg">
    <line/>
</svg>
```
It tries to access position attributes that dont exist.

### LINKED ISSUE #729: Attributes get wrongly added, as if user created new tag
### MRP
```svg
<svg width="8" height="8" xmlns="http://www.w3.org/2000/svg">
    <circle/>
</svg>
```
will turn into
```svg
<svg width="8" height="8" xmlns="http://www.w3.org/2000/svg">
    <circle r="1"/>
</svg>
```

### Proposed solution
Separate `_init()` from user added tag setup.

### Full issue
All supported tags:
```svg
<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg">
    <circle/>
    <rect/>
    <line/>
    <ellipse/>
    <path/>
</svg>
```
currently turns into:
```svg
<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg">
    <circle r="1"/>
    <rect width="1" height="1"/>
    <line/>
    <ellipse rx="1" ry="1"/>
    <path d="M0 0"/>
</svg>
```

Kiisu-Master:
I separated setting up tag specific defaults in to method `user_setup()` so it doesn't add new attributes to tags that don't have any. The attribute values now only get changed when adding new tags with the UI actions.
Fix #729

I also changed the initialization of attributes to fix some not being initialized. This might change the order of attributes in the UI, but that can be easily fixed.
Fix #726

### DIFF
--- a/src/data_classes/Tag.gd
+++ b/src/data_classes/Tag.gd
@@ -17,6 +17,10 @@ func _init() -> void:
 	for attribute: Attribute in attributes.values():
 		attribute.propagate_value_changed.connect(emit_attribute_changed)
 
+func user_setup(_what = null) -> void:
+	printerr("Unimplemented function")
+	breakpoint
+
 func set_unknown_attributes(attribs: Array[Attribute]) -> void:
 	unknown_attributes = attribs.duplicate()
 	for attribute in unknown_attributes:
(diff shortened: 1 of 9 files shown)


Kiisu-Master:
Theres something wrong with attribute change propagation now

MewPurPur:
Seems like a good solution for now.

Thanks!

PR #740, opened 2024-05-17
Add ColorDB Singleton

Qainguin:
This pull request adds a ColorDB Singleton to the project and replaces all mentions from ThemeGenerator colors to ColorDB. The reason for this is so I (or others) may create a way to make custom editor themes for GodSVG.


Qainguin:
Also please disregard the first four commits, they failed in a lot of ways. I needed to make this ColorDB first before making themes so I figured why not make a pull request as soon as I could.

Kiisu-Master:
Nice commits you have here

MewPurPur:
Nice idea! Custom themes are something GodSVG should support. Currently, the top item on the "support themes" checklist is reducing the number of theme colors a little, not needing to define so many of them for a theme.

My big reservation is that I'm not sure this is the right infrastructural change for multiple themes to happen. A change like this should happen either along with the full implementation, or after working out full implementation details.

Qainguin:
With that, I think another one of the biggest problems is how themes are set up. Currently, it's just one big function that updates everything but I would like to be able to update individual themes without updating everything.

Qainguin:
Actually, I could do something with having the one base color and then an "Advanced Options" setting that gives you finer control.

MewPurPur:
> With that, I think another one of the biggest problems is how themes are set up. Currently, it's just one big function that updates everything but I would like to be able to update individual themes without updating everything.

Does it matter? If your concern is performance, the setup isn't expensive, it's the merging. Maybe I'm wrong, but I'd need to see benchmarks proving it.

> Also, we could probably get one base color for a theme and then build everything else off of that one color.

Yeah, but normally you need at least 2 colors for a theme. And I think the theme generator is kind of misleading with the colors, there would probably be different setups too, like low/medium/high "roundiness" for example.

Qainguin:
> Does it matter? If your concern is performance, the setup isn't expensive, it's the merging. Maybe I'm wrong, but I'd need to see benchmarks proving it.
When I tested updating every time the color was changed there was a lag spike which feels less than ideal.

MewPurPur:
When I tested it, creating the theme took something like 3ms, merging it with the main theme took way more. The rest is the cost of redrawing the UI, which mostly gets high when you have a big SVG.

PR #743, opened 2024-05-19
Fix the program crashing when confirming an invalid setting value

### LINKED ISSUE #669: Crash when confirming empty setting
GodSVG will crash because of NAN value when having empty settings for ui scale or handle size.

ilikefrogs101:
Fixes #669

### DIFF
--- a/src/parsers/NumberParser.gd
+++ b/src/parsers/NumberParser.gd
@@ -18,7 +18,7 @@ static func format_text(text: String) -> String:
 		return ""  # Equivalent to NAN in the app's logic.
 	
 	var leading_decimal_point := text.begins_with(".") or text.begins_with("-.") or\
-			text.begins_with("+.")
+		text.begins_with("+.")
 	var padded_zeros := 0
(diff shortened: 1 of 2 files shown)


src/parsers/NumberParser.gd
```diff
@@ -50,17 +50,17 @@ static func evaluate(text: String) -> float:
 	var expr := Expression.new()
 	var err := expr.parse(text.replace(",", "."))
 	if err == OK:
-		var result: float = expr.execute()
-		if not expr.has_execute_failed():
-			return result
+		var result: String = var_to_str(expr.execute())
+		if not expr.has_execute_failed() and result.is_valid_float():
+			return str_to_var(result);
```
MewPurPur, line 55: Let's not have semicolons.

ilikefrogs101: Oops, lmao. every single time. I cant avoid it apparently

ilikefrogs101: Fixed

MewPurPur:
Thanks!

PR #750, opened 2024-05-20
Fix translated shortcuts

MewPurPur:
Fixes the "Delete" shortcut name auto-translating. Temporary fix for alpha 4, before we move to 4.3

Tweaks the squircle icon to have the MacOS level of roundiness. (probably? Where the fuck is the guideline that everyone seems to know about?)


visual/squircle_icon.svg
Qainguin, line 1: It feels like the fox should be smaller in this.

PR #777, opened 2024-05-31
russian translation cleanup

JinEnMok:
Fixed a few errors in the existing translation and added a few to items that were previously not translated. 

There are still things that are not translated, e.g. the "width" and "height" fields, some buttons in the keyboard shortcut submenu, the right-click menu, etc.

### DIFF
--- a/translations/ru.po
+++ b/translations/ru.po
@@ -1,27 +1,26 @@
-#, fuzzy
 msgid ""
 msgstr ""
 "Project-Id-Version: GodSVG\n"
 "POT-Creation-Date: \n"
-"PO-Revision-Date: \n"
-"Last-Translator: volkov <volkovissocool@gmail.com>\n"
-"Language-Team: \n"
+"PO-Revision-Date: 2024-05-31 11:16+0300\n"
+"Last-Translator: Gallifreyan <gallifreyan@protonmail.ch>\n"
 "Language: ru\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"X-Generator: Poedit 3.4.2\n"
+"X-Generator: Lokalize 24.05.0\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 msgid "translation-credits"
 msgstr "volkov <volkovissocool@gmail.com>"
 
 #: src/parsers/SVGParser.gd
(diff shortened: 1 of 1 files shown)


translations/ru.po
```diff
@@ -87,7 +87,7 @@ msgstr "Подвинуть вниз"
 
 #: src/ui_elements/palette_config.gd
 msgid "Copy as XML"
-msgstr "Скопировать как XML"
+msgstr "Скопировать XML"
```
MewPurPur, line 90: Are you sure about this? The context for this is, you're converting a ColorPalette resource into XML and copying that, so you can easily share it around. There's no XML until you "copy as XML".

translations/ru.po
```diff
@@ -239,27 +239,27 @@ msgstr "Внешний вид"
 
 #: src/ui_parts/display.gd src/TranslationUtils.gd
 msgid "Load reference image"
-msgstr ""
+msgstr "Загрузить "
```
MewPurPur, line 242: I think this is incomplete?

translations/ru.po
```diff
@@ -3,25 +3,25 @@ msgid ""
...
 "Language: ru\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"X-Generator: Poedit 3.4.2\n"
+"X-Generator: Lokalize 24.05.0\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 msgid "translation-credits"
 msgstr "volkov <volkovissocool@gmail.com>"
```
MewPurPur, line 16: Add yourself if you want to be credited

JinEnMok: do I add another msgid/msgstr or do i append myself to the existing msgstr?

translations/ru.po
JinEnMok: Good catch, cheers

translations/ru.po
MewPurPur: I don't remember, look at the German localization file.

MewPurPur:
I'm preparing alpha 4 right now, but if you're quick, I can sneak this in just before I bump the version and cease all development until the release. (* ^ ω ^)

JinEnMok:
Should be done now. Will look into translating non-translated fields for the next release then

MewPurPur:
Thanks!

PR #786, opened 2024-06-06
Rework Reference Images

Qainguin:
- Now saves the selected reference image to "user://reference_image.png"
- Removed reference path from SaveData as it is not needed anymore

Only changes things in the backend, nothing is changed for the user. This should make implementing web support easier.


MewPurPur:
Thanks!

PR #799, opened 2024-07-03
Core infrastructure rework & Implement group and gradient tags

### LINKED ISSUE #56: Support <linarGradient> natively
An initial implementation should have the following:

- Implement IDs.
- Add relevant attributes of \<linearGradient> and \<stop>.
- Add gradients in the color popup, next to the "none" color.

Would be great to also implement some kind of visual for how the gradient looks.

### LINKED ISSUE #434: Support the <g> tag natively
I think for an initial implementation, the following things would be needed:

- Add a bunch of attributes to it, like transform, stroke, fill, etc.
- Make it so attributes update their "default" values based on the <g> ancestor tags.
- Make it so transformations applied to the <g> tag are applied to its child tags.

This the biggest remaining thing that needs to be implemented before GodSVG can get out of alpha.

### LINKED ISSUE #801: Focusing the elliptical arc radius parameter field when it's negative changes it
https://github.com/user-attachments/assets/af92ec3d-94fc-46f0-a4e3-ed9a8979f0c9

MewPurPur:
This isn't close to being done, but now is a decent time to post it.

The bulk of this is a rework to how GodSVG's core infrastructure works, though there are some additional functionalities I got onto while doing this stuff.

- AppInfo.gd is now app_info.json (as it's advantageous to have this be easily fetched from other places, like the GodSVG website)
- Issue templates on Github are now .yaml for a better experience
- Implemented "g", "linearGradient", "radialGradient" tags (Resolves #56, Resolves #434)
- DB now holds the propagatability of all known attributes.
- Tags now correctly consider the attribute defaults established by the `<svg>` tag and underlying `<g>` tags
- Tags now correctly calculate the accumulated transform from underlying `<g>` tags
- Tags now hold their own defaults for non-propagatable attributes (as this is an important aspect of the SVG spec). This is necessary to make some attributes with "auto" values not misleading
- DB now holds the types of all known attributes, so that even in unknown tags, they can give utilities (e.g. `fill` inside a `<polygon>` tag will now be recognized as a color attribute)
- Attribute defaults are now tag-specific
- Added TagRoot for the root svg, and TagSVG which may in the future be used for nested SVGs
- TagRoot no longer needs to self-replace on unknown changes, now the SVG singleton emits the same signals and can be connected to reliably
- Added functionality for tag configuration warnings (e.g. `<stop>` tags show a configuration warning when placed under anything other than a gradient)
- Added functionality for shape bounding boxes, will be useful in the future for transformation gizmos and maybe other things
- Added XNode class that Tag inherits from, will be useful in the future to avoid destroying other XML nodes (i.e. comment, CDATA, text)
- Implemented attributes of type ID and a corresponding field
- Implemented rudimentary pretty formatting. Sorry for taking so long - I wanted to do it better than this, but I hope having it be just a checkbox without any configurability will be fine for now
- Tweaked some colors. Changed yellow movement indicators to green, to indicate when the operation is right. Orange might be used in the future for invalid operations (e.g., moving a `<stop>` tag under something that's not a gradient)
- Used the new "Swatch" theme variation for the fake swatch in the palette editor
- Adds a SHORTCUTS_CHANGED notification (Supersedes #671)
- Fixes to the order of drawing in the viewport
- Import Warnings dialog no longer repeats itself
- Added completion percentage to the Language button
- Changed all of the app's icons to consistently use width and height and no viewBox
- Added autoformatting setting for not applying autoformatting to pathdata and transform list attributes. Of course, it only holds for typing them out
- Attributes of the root tag are now added to the UI
- Implemented a setting for window title scheme. When I record GodSVG, the window titles would throw off OBS, so at least to me, this will be useful
- Fixes #801
- Reworked UndoRedo system to be more robust
- Made tag conversion more reliable
- Renamed tags to elements
- Added support for percentages


Qainguin:
I just downloaded the Action run for Windows and it just crashes on launch. I already cleared all saved files in the appdata folder.

MewPurPur:
There's more work to be done, can't guarantee that this is stable until I merge it (or even after, I expect some regressions)

Nonetheless thanks for testing! If you know how to run it through a console and get the error message from there, it'd be really helpful.

MewPurPur:
@Qainguin The PR is more stable now, I'll be doing more testing over the next week before pushing. Could you check again?

PR #803, opened 2024-07-22
Fix Ctrl + Middle mouse drag zooming

Kiisu-Master:
Due to a hardware difference i was not aware of this earlier. On my laptop the keyboard probably doesn't spam input when holding a key down.


MewPurPur:
Thanks!

PR #814, opened 2024-07-23
Add a CI check for file formatting

aaronfranke:
This PR adds a CI script that ensures proper POSIX text file formatting, no trailing whitespace, and a few other things. I excluded SVG files from this check, but it will run for most text files (GDScript, markdown, etc).


.github/file_format.sh
```diff
@@ -0,0 +1,53 @@
...
+while IFS= read -rd '' f; do
+    # Exclude some types of files.
+    if [[ "$f" == *"svg" ]]; then
+        continue
+    fi
+    # Ensure that files are UTF-8 formatted.
+    recode UTF-8 "$f" 2> /dev/null
+    # Ensure that files have LF line endings and do not contain a BOM.
+    dos2unix "$f" 2> /dev/null
+    # Remove trailing space characters and ensures that files end
```
MewPurPur, line 19: Grammar typo

MewPurPur:
I'm fine with this, but I'd prefer that we don't trim tab-only lines. I think having empty lines among indented ones doesn't look nice, especially in the Godot editor. It's also inconvenient for me, forcing me to re-indent if I want to type on them.

MewPurPur:
Thanks!

PR #820, opened 2024-07-24
Fix workflows running on wrong Godot version

Kiisu-Master:
Changed it back to use version and release separately.
Also had to update the way wine and rcedit get added to the editor settings because the file names changed to include godot version.
Also executable sizes even larger now 💀


Kiisu-Master:
Should be ready now

Kiisu-Master:
Actually wait this might break on patch releases again because of the file name of the editor settings only includes main versions

MewPurPur:
Thanks!

PR #851, opened 2024-08-03
Rework the settings menu

### LINKED ISSUE #847: ".%" numeric attribute value crashes GodSVG
`<svg xmlns="http://www.w3.org/2000/svg" width="8" height="8"><rect x=".%"/></svg>` This trivial SVG causes a crash.

### LINKED ISSUE #850: Rethinking attribute formatting
The general setup for formatting has to take into account the following methods:
- Enum and ID attributes can only be modified as text, and shouldn't be formatted
- Numeric attributes can be modified as text or via setting the number directly
- Color attributes can be modified as text or via the color picker
- Transform list attributes can be modified as text or via the transform editor
- Pathdata attributes can be modified as text or via the path commands editor

So outside of text, numeric attributes just generate a float, the color picker generates a 6-letter hex code, the transform editor generates an array of transforms, and the path editor generates a an array of path commands. And that is all pretty rigid, no personal style taste, and gets formatted automatically into text according to formatting rules.

For the 4 things that can currently be edited as text, I think attribute formatting from now on should be all about how the underlying representations are converted into text, and not the full cycle.

# New idea

## XML formatting

First of all, the XML formatting section. This is separate from attribute formatting and will concern the SVGParser class. This could for now involve the following formatting settings:

**Add newline at the end (off by default)**

**Use shorthand tags: Enabled / Only self-contained elements (default) / Disabled**
- Note: Self-contained elements are ones that don't expect to have children: Currently the basic shapes and the stop element.

**Pretty XML structuring (off by default)**

**Indentation: Tabs (default) / Spaces**
- Only active when pretty XML structuring is on.

**Space indentation size (2 / 3 / 4 / 6 / 8, but allows any up to 20)**
- Only active when pretty XML structuring is on and indentation is spaces.

## Universal attribute formatting

All attributes should be formatted in at least one way: the edges must be trimmed to remove spaces, tabs, and newlines. This includes attributes that aren't formatted, like enums.

Aside from this, all of the formatted attribute types would have a "Format manually inputted text" option that's on by default. If the attribute that has been inputted in text form is valid, no interference needed, then when this setting is disabled, it won't be formatted.

Numbers are an exception, as they may be restricted by numeric bounds and apply numeric precision unto itself, and as such require to care about the numeric representation even when you're inputting text

### Attribute formatting

There would be prepared presets for certain kinds of use cases:

- Most compact: Maximum compactness settings
- Compact: Disables some of the more sporadic compactness settings
- Comfortable (default)
- Pretty: Enables some aesthetic settings that take more space
- Custom

### Numeric attributes:

**Remove leading zero (off by default)**
- 0.5 > ".5"
- -0.8 > "-.8"

### Color attributes

**Use named colors: Always / Only if shorter or equal / Only if shorter / Never**

**Alternative format: 6-digit or 3-digit hex / 6-digit hex / RGB**
- RGB example: "#ff0052" > "rgb(255, 0, 82)"
- 6 or 3-digit hex example: "#dd0055" > "#d05"

-----

Paths and transforms are already pretty nicely implemented and don't need to change.

MewPurPur:
- Changes the mess that is the settings menu into a different mess that can do more things
- Implements the ability to define multiple formatters.
- Resolves #850


src/config_classes/Formatter.gd
```diff
@@ -0,0 +1,149 @@
...
+func _init(new_title := "") -> void:
+	title = new_title
+	changed.connect(_on_changed)
+
+func modify_title(new_title: String) -> void:
+	title = new_title
+	emit_changed()
+
+func _on_changed() -> void:
+	GlobalSettings.save()
```
Swarkin, line 147: You're saving the GlobalSettings while they are being parsed here, resulting in all the weird errors.

Swarkin: This also means that the earlier ConfigFile implementation was perfectly fine by itself, and you _could_ make a pr / issue to get that system back, as it is a tiny bit easier to be edited by another human than Resource files, in my opinion

src/config_classes/Formatter.gd
```diff
@@ -0,0 +1,150 @@
...
+
+
+func _init(new_title := "") -> void:
+	title = new_title
+	changed.connect(_on_changed, CONNECT_DEFERRED)
+
+func _on_changed() -> void:
+	if self == GlobalSettings.savedata.editor_formatter:
+		SVG.sync_elements()
+	GlobalSettings.save()
```
Swarkin, line 150:
```suggestion
func _init(new_title := "") -> void:
	title = new_title
	# Connect to _workaround after one frame (CONNECT_DEFERRED) and only once (CONNECT_ONE_SHOT)
	changed.connect(_workaround, CONNECT_DEFERRED | CONNECT_ONE_SHOT)

func _workaround() -> void:
	# Connects to the _on_changed function once all initial file loading is done.
	# This avoids saving the config 24 times on launch, during initialization and loading.
	changed.connect(_on_changed)

func _on_changed() -> void:
	if self == GlobalSettings.savedata.editor_formatter:
		SVG.sync_elements()
	GlobalSettings.save()

```

PR #862, opened 2024-08-18
Update Russian and Ukrainian translations

Vovkiv:
Finished now, but didn't tested that much except quickly glancing


translations/ru.po
```diff
@@ -84,63 +84,62 @@ msgid "Insert After"
 msgstr "Вставить после"
 
 #: src/config_classes/Formatter.gd
-#, fuzzy
 msgid "Always"
-msgstr "Всегда активно."
+msgstr "Всегда активно"
```
MewPurPur, line 88: It shouldn't be "always active". That string is retired, this is in the context of "Always / Sometimes / Never" sorts of dropdowns.

translations/ru.po
```diff
@@ -84,63 +84,62 @@ msgid "Insert After"
...
 
 #: src/data_classes/ElementG.gd
 msgid "This group has only one element."
-msgstr ""
+msgstr "В этой группе только один элемент."
 
 #: src/data_classes/ElementLinearGradient.gd
 msgid "No id attribute defined."
-msgstr ""
+msgstr "Атрибут идентефикатора не определен."
```
MewPurPur, line 128: "id" is the name of the attribute.

MewPurPur: Oh I see, yeah this string has a version with quotation marks too. That one is valid, and I'll erase this one.

translations/ru.po
Vovkiv: I guess I just missed string haha

MewPurPur:
Any update? I partially finished translation is also fine, it's still better than none.

MewPurPur:
Thanks!

PR #882, opened 2024-08-26
Github actions time

MewPurPur:


Kiisu-Master:
The set up godot thing wont work on mac os because that has different locations for files. Also why did you change it to "setup" godot, I think thats incorrect english

PR #909, opened 2024-09-10
Tweak web workflow

MewPurPur:
Tweaks the web workflow to give the zip instead of the export template like for all other platforms.

### DIFF
--- a/.github/workflows/export-optimized.yml
+++ b/.github/workflows/export-optimized.yml
@@ -209,9 +209,18 @@ jobs:
           if-no-files-found: error
           retention-days: 28
   build-web:
-    name: Get GodSVG export template for Web
+    name: Export GodSVG for Web
     runs-on: ubuntu-latest
     steps:
+      - name: Set up Godot Editor
+        run: |
+          mkdir -v -p ~/godot-editor
+          cd ~/godot-editor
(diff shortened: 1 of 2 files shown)


.github/workflows/export-optimized.yml
```diff
@@ -242,12 +251,20 @@ jobs:
         run: |
           cd godot
           source ../emsdk/emsdk_env.sh
-          scons p=web arch=wasm32 optimize=size lto=auto threads=no ${BUILD_OPTIONS}
+          scons p=web arch=wasm32 ${BUILD_OPTIONS} threads=no
+          mkdir -v -p ~/.local/share/godot/export_templates/${GODOT_VERSION}.${GODOT_RELEASE}/
+          mv ./bin/godot.web.template_release.wasm32.nothreads.zip ~/.local/share/godot/export_templates/${GODOT_VERSION}.${GODOT_RELEASE}/web_release.zip
```
Kiisu-Master, line 256:
```suggestion
          mv ./bin/godot.web.template_release.wasm32.nothreads.zip ~/.local/share/godot/export_templates/${GODOT_VERSION}.${GODOT_RELEASE}/web_nothreads_release.zip
```

Kiisu-Master: also remove this wrong path https://github.com/MewPurPur/GodSVG/blob/main/export_presets.cfg#L386

Kiisu-Master (changes requested):

.github/workflows/export-optimized.yml
MewPurPur: Thanks!

PR #972, opened 2024-11-17
Web file rework

### LINKED ISSUE #752: Implement Web Support for Reference Images
Add support for reference images for web import. Currently, it hides the reference image button for web users, which is more of a band-aid than anything.

MewPurPur:
Reworks how files are handled on web. Closes #752


Qainguin:
Very good!

PR #1001, opened 2024-12-01
Android: Fix viewport zoom issue

syntaxerror247:
- Fixes https://github.com/MewPurPur/GodSVG/issues/160

The issue was that the Pan gesture was interfering with the Magnify gesture, and I couldn't identify a specific use case for the Pan gesture. Therefore, I have disabled it for Android. I chose not to remove the code entirely, as its exact usage is unclear. 

Everything is working as expected in my tests.


MewPurPur:
Merging in spite of the codebase freeze for the upcoming alpha 6, since this only changes behavior on a platform that's currently not exported.

MewPurPur:
Thanks!

PR #1003, opened 2024-12-02
Unify X11 and Wayland again

MewPurPur:
It would appear that Godot tries out all display servers it can in order, and X11 and Wayland are both available by default, so there's no need for separate builds. Changes the default to try out Wayland first (even though it's experimental).


MewPurPur:
Hmm... I feel confident enough in this.

PR #1010, opened 2024-12-04
Angle support for Windows

syntaxerror247:


MewPurPur:
Looks good! I won't merge this yet because I'm just concerned about adding 4 MB worth of files to the repository, especially with how it looks like we may add more of these in the future. But if worst comes to worst, we'll go with it.

Issue #793, opened 2024-06-21
Moving your mouse a little while clicking on the tag button makes it unpress.


MewPurPur:
![image](https://github.com/MewPurPur/GodSVG/assets/85438892/6ab9a37a-5383-42f8-ba41-dafc98103a10)

These buttons have the issue described in the title.

Swarkin:
this is usually considered to be a feature, so that you can move away the mouse to cancel the clicking operation

however, most of the time you have to move the mouse away from the button's entire bounding box instead of just a single pixel

MewPurPur:
Yes, exactly.

PR #1019, opened 2024-12-08
Fix element frame button deselecting

syntaxerror247:
- Fixes https://github.com/MewPurPur/GodSVG/issues/793

To prevent dragging when interacting with the `title_button`, its `mouse_filter` was set to STOP. This caused the `title_button` to capture the mouse input, triggering the `mouse_exited` signal of the base node. The `mouse_exited` signal deleted the current `title_button`, which allowed the base node to capture the mouse input again. Subsequently, the mouse_entered signal of the base node was triggered, creating a `new title_button`. However, this new button was unpressed due to the transition.

The suggested solution introduces a flag to suppress dragging while the title_button is pressed. This prevents the dragging without unintentionally deleting and recreating the button.


MewPurPur:
Thanks!

PR #1021, opened 2024-12-10
Android: Increase default handle size

syntaxerror247:
This PR introduces a new setting that automatically adjusts the handle size based on the screen size. This setting is enabled by default, ensuring that the selection handles are appropriately scaled.

### DIFF
--- a/src/autoload/GlobalSettings.gd
+++ b/src/autoload/GlobalSettings.gd
@@ -256,6 +256,10 @@ func reset_settings() -> void:
 	modify_setting("formatters", formatters_array, true)
 	modify_setting("editor_formatter", pretty_formatter, true)
 	modify_setting("export_formatter", compact_formatter, true)
+	
+	# Higher default handle_size on Android for better touch control.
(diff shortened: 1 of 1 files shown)


MewPurPur:
But aren't handle sizes already relative to other UI elements? Could you show better what the PR accomplishes? I couldn't figure out anything in testing.

syntaxerror247:
This is on my laptop:

https://github.com/user-attachments/assets/ffad466a-f335-4c43-9eb3-6b941bc97d4a

Difference is better visible on smaller screen size, like Android:

https://github.com/user-attachments/assets/e1565c3e-71e9-4c85-adcd-2c634b52c5c5

MewPurPur:
So you want handles to be scaled up using the same thing that currently determines the UI scale?

That's not a good way. Currently, handles are always the same size relative to other UI elements, times a multiplier. If you have small handles, then you have a small UI too, so the problem is not inherently in the handle size. If the issue is that small handles are hard to deal with on touchscreens, you should solve it in a way that would also affect Desktop.

If the idea

MewPurPur:
Thanks!

PR #1022, opened 2024-12-11
Implement Floating Shortcut Panel for Android

syntaxerror247:
This PR implements a **Shortcut Panel** for Android. The toolbar floats above the UI, with buttons that trigger key events (e.g., undo and redo actions).
- It includes a drag handle for repositioning. 
- Initially only undo and redo buttons are available. However it gives an option to edit and add extra shortcut button.


MewPurPur:
I really like it!

There needs to be something to prevent you from having the same setting multiple times.

MewPurPur:
We should also not forget to translate the shortcuts

syntaxerror247:
> There needs to be something to prevent you from having the same setting multiple times.

Restricted assigning the same shortcut to multiple slots and added an advice panel.

syntaxerror247:
@MewPurPur I updated the shortcut selection panel to use the shortcut description instead of the shortcut name and aligned it to the left. Also, updated the translation files.

MewPurPur:
Clear button is needed.

Used shortcuts should be disabled in the popup and also highlighted in red when being typed.

The new icons should use the "#def" fill color which is standard for GodSVG.

The config option shouldn't be the gear with a black outline - that one is used in color swatches because they can have arbitrary colors, but is not appropriate here.

MewPurPur:
Now it's pretty good. I think the options should be disabled rather than removed from the popup.

Now there's the issue that the panel gets hidden when you open menus. We can leave it for later.

Make the dragging icon bigger to match its actual size and avoid upscaling artifacts (I'd suggest having 4x2 dots also)

syntaxerror247:
@MewPurPur Options are now disabled instead of being removed entirely. I also increased the icon size and eliminated upscaling.

MewPurPur:
Okay, last set of requests before I look at the code:

The panel needs to have better padding.

You shouldn't be able to drag it too out of bounds.

visual/icons/DragHandle.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="20" height="16" xmlns="http://www.w3.org/2000/svg">
+	<g fill="#def">
+		<circle cx="3" cy="4" r="2" />
+		<circle cx="10" cy="4" r="2" />
+		<circle cx="17" cy="4" r="2" />
+		<circle cx="3" cy="12" r="2" />
+		<circle cx="10" cy="12" r="2" />
+		<circle cx="17" cy="12" r="2" />
+	</g>
+</svg>
```
MewPurPur, line 10:
```suggestion
<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg"><g fill="#def"><circle cx="4" cy="4" r="2"/><circle cx="12" cy="4" r="2"/><circle cx="12" cy="20" r="2"/><circle cx="4" cy="12" r="2"/><circle cx="4" cy="20" r="2"/><circle cx="12" cy="12" r="2"/></g></svg>
```

I'd prefer that we use an SVG like this. I've explained in other places how I'd like for it to be used. Also, "handle" has other uses within the codebase, so name this file GripDots, GripPattern, GrabberDots, something like that.

src/ui_parts/shortcut_panel.gd
```diff
@@ -0,0 +1,56 @@
...
+func simulate_key_press(action_name: String) -> void:
+	var events := InputMap.action_get_events(action_name)
+	for event in events:
+		if event is InputEventKey:
+			event.pressed = true
+			Input.parse_input_event(event)
+			return
+
+func _on_panel_settings_pressed() -> void:
+	get_tree().root.add_child(PanelSettings.instantiate())
```
MewPurPur, line 56: Use HandlerGUI.add_menu()

src/ui_parts/shortcut_panel.tscn
```diff
@@ -0,0 +1,57 @@
...
+layout_mode = 2
+theme_override_constants/separation = 8
+alignment = 1
+
+[node name="DragHandle" type="TextureRect" parent="PanelContainer/HBoxContainer"]
+custom_minimum_size = Vector2(35, 0)
+layout_mode = 2
+mouse_default_cursor_shape = 6
+texture = ExtResource("2_8txvx")
+stretch_mode = 3
```
MewPurPur, line 38: I think we should have a structure like this:

PanelContainer
- Dots
- MarginContainer
  - HBox
    - Buttons

The panel container could be edited from its theme variations to have smaller paddings (look for example at main_scene.tscn the update_theme() method). On top, left, and right of the buttons, the padding should be the same.

src/ui_parts/shortcut_panel.tscn
```diff
@@ -0,0 +1,57 @@
...
+theme_override_constants/h_separation = 4
+icon = ExtResource("3_lpgyc")
+icon_alignment = 1
+
+[node name="HBoxContainer" type="HBoxContainer" parent="PanelContainer/HBoxContainer"]
+layout_mode = 2
+theme_override_constants/separation = 8
+
+[node name="spacer" type="Control" parent="PanelContainer/HBoxContainer"]
+layout_mode = 2
```
MewPurPur, line 55: See the previous review comment for how I expect the padding to be achieved.

src/ui_parts/shortcut_panel.tscn
```diff
@@ -0,0 +1,57 @@
...
+
+[node name="PanelSettings" type="Button" parent="PanelContainer/HBoxContainer"]
+custom_minimum_size = Vector2(30, 30)
+layout_mode = 2
+size_flags_vertical = 4
+focus_mode = 0
+mouse_default_cursor_shape = 2
+theme_override_constants/h_separation = 4
+icon = ExtResource("3_lpgyc")
+icon_alignment = 1
```
MewPurPur, line 48: Make this button themed differently. TranslucentButton theme variation should be good. I also made an icon I think would be better for a "configure" kind of option, as to disassociate it from settings.

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><path fill="#def" d="M4 1 3.5 3l.5 2v3H2.5a.5.5 0 00-.5.5v5.5a1 1 0 001 1h3a1 1 0 001-1v-5.5a.5.5 0 00-.5-.5H5V5l.5-2-.5-2zM4 9v5h-1v-5zM6 9v5h-1v-5zM10 1.174a3 3 0 000 5.652V14a1 1 0 002 0V6.824a3 3 0 000-5.648V4l-1 .5-1-.5z"/><path fill="#def" d="M3 9h3v5h-3z" opacity=".6"/></svg>
```

syntaxerror247: why no minimum size?

MewPurPur: The minimum size should be determined by the icon and stylebox for simplicity.

src/ui_parts/shortcut_panel.gd
syntaxerror247: Add menu will also add a color rect, which will stop the input to editor.

MewPurPur:
Thanks!

PR #1027, opened 2024-12-13
Implement TitledPanel container

MewPurPur:
Implements a TitledPanel container which is a container that has a "title" part and a "panel" part under it represented by two child nodes. This is just a chunk of custom UI that does the role of 3 nodes in 1, and does it better, although in a more limiting way.

So it's cheaper and better in those situations.

Later on I'll look into adding some more features to this, so it can be used to replace this UI pattern everywhere it appears.


syntaxerror247:
Thanks

MewPurPur:
Whuh?

PR #1029, opened 2024-12-17
CI: Update Android build

syntaxerror247:


MewPurPur:
I don't want my PRs to trigger workflows that are any different from those of other contributors. Would it be okay to split the Android release as a separate workflow that needs to be manually dispatched?

syntaxerror247:
> I don't want my PRs to trigger workflows that are any different from those of other contributors. Would it be okay to split the Android release as a separate workflow that needs to be manually dispatched?

This is possible to create a seperate workflow but I don't see any problem doing it this way. It'll just run one extra job that will be for release build.

MewPurPur:
Thanks!

PR #1031, opened 2024-12-25
Improve Chinese (Simplified) translation

mikelei8291:
Translated all missing strings and reviewed existing ones.


MewPurPur:
Thanks!

PR #1039, opened 2024-12-30
Temp fix for UI scaling on Android

syntaxerror247:


MewPurPur:
I think this may be a common problem with small screens or high DPI displays, but as you've said, we can look into it later.

MewPurPur:
Thanks!

PR #1040, opened 2024-12-30
CI: Set Android build debug keystore

syntaxerror247:
This change configures the project to use a single, consistent debug keystore, enabling the installation of testing builds without the need to uninstall previous versions of the app. This ensures a smoother testing process by maintaining the same keystore across builds, allowing for easier updates and debugging


MewPurPur:
Thanks!

PR #1042, opened 2024-12-30
Add scrollbar support for Dropdown

syntaxerror247:


MewPurPur:
I should investigate if min_width even matters that much in our UI. The scrollbar setting should be max_height, not minimum as that doesn't make sense - so I think it should be a separate int either way.

syntaxerror247:
> I should investigate if min_width even matters that much in our UI. The scrollbar should show after exceeding a max_height, not minimum, as that doesn't make sense - so I think it should be a separate int either way.

I think there might be some confusion about the terms. When scrolling is enabled in a ScrollContainer, we need to set its size manually; otherwise, it defaults to the smallest size possible. That's why I'm setting its minimum height, in this context, it effectively acts as a fixed height since it won't grow beyond that.

src/ui_parts/about_menu.tscn
```diff
@@ -474,4 +475,7 @@ focus_mode = 0
 mouse_default_cursor_shape = 2
 text = "Close"
 
+[node name="Dropdown" parent="." instance=ExtResource("7_xb47m")]
+layout_mode = 2
+
```
MewPurPur, line 480: Seems like a mistake?

src/ui_widgets/ContextPopup.gd
```diff
@@ -217,11 +220,30 @@ min_width := -1.0, separator_indices := PackedInt32Array()) -> void:
...
+	scroll_container.horizontal_scroll_mode = ScrollContainer.SCROLL_MODE_DISABLED
+	scroll_container.vertical_scroll_mode = ScrollContainer.SCROLL_MODE_DISABLED
+	
+	# Increase the scrollbar area on Android.
+	if OS.get_name() == "Android":
+		var scrollbar := scroll_container.get_v_scroll_bar()
+		var stylebox := scrollbar.get_theme_stylebox("scroll").duplicate()
+		stylebox.content_margin_left = 10
+		stylebox.content_margin_right = 10
+		scrollbar.add_theme_stylebox_override("scroll", stylebox)
```
MewPurPur, line 241: I think it would make more sense to add this as a theme variation inside ThemeUtils, or even have a condition for the default scroll styleboxes to make the content margin bigger on Android.

src/ui_widgets/dropdown.gd
```diff
@@ -5,6 +5,8 @@ extends HBoxContainer
 
 @export var values: PackedStringArray
 @export var restricted := true
+## If set to a value greater than 0, the popup will have a fixed height, and scrolling will activate if the content exceeds it.
+@export var popup_panel_height := 0
```
MewPurPur, line 9: Have it at -1.0 like the default value of the argument.

Also, is there actually any reason to have width/height be floats?

src/ui_parts/about_menu.tscn
syntaxerror247: Yes, I was testing something earlier and forgot to clean it up.

src/ui_widgets/ContextPopup.gd
syntaxerror247: The scrollbar is only needed for these popup panels, as they contain buttons, making touch scrolling support tricky. Everywhere else, touch scrolling works perfectly fine. That's why I made a direct modification here.

src/ui_widgets/dropdown.gd
syntaxerror247: > Also, is there actually any reason to have width/height be floats?

For the `min_size`, we're directly using a value like `size.x`, which is a float. When converted to an integer, this causes a loss of precision. However, the impact on our use case will be negligible, and it will only result in warnings in the editor `Narrowing conversion: float is converted to int and loses precision` if we don't manually handle the conversion.

I would suggest keeping it as a float.

src/ui_widgets/ContextPopup.gd
MewPurPur: Hmm... okay, we can do that if the pattern appears again.

MewPurPur:
Thanks!

PR #1048, opened 2025-01-04
Rework SaveData

MewPurPur:
Cleans up the system around SaveData.

- Implements ConfigResource class which implements the useful `changed_deferred` signal. This signal is used in situations where the many times `changed` can run aren't necessary, and prevents us from instead needing to implement queueing in many different places.

- Moves SaveData operations from the Configs singleton to the resource itself. Setters in SaveData now emit the singleton signals and call modification callbacks, meaning that things should be modified directly in the SaveData now.

- Some exported properties in the config resources are now private and can only be interacted with through methods.

- Shortcuts are now an empty dictionary, only modifications are stored. Logic is changed to work around that.

- Validations now also happen within SaveData. All settings with defaults are reset in reset_config(), as well as other initial settings. There is universal validation of the editor and export formatters, as a savefile not having them would make GodSVG impossible to use.

- Recent dirs are now an array with 5 elements, which will later be used for the custom file dialog and possibly other things.

### DIFF
--- a/src/config_classes/SaveData.gd
+++ b/src/config_classes/SaveData.gd
@@ -1,7 +1,11 @@
-class_name SaveData extends Resource
+class_name SaveData extends ConfigResource
 
 const GoodColorPicker = preload("res://src/ui_widgets/good_color_picker.gd")
 
+var _palette_validities := {}
+var _shortcut_validities := {}
+
+# Most settings don't need a default.
 func get_setting_default(setting: String) -> Variant:
 	match setting:
 		"highlighting_symbol_color": return Color("abc9ff")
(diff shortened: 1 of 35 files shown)


src/config_classes/SaveData.gd
```diff
@@ -32,58 +36,482 @@ func get_setting_default(setting: String) -> Variant:
...
+			emit_changed()
+			export_formatter.changed.connect(emit_changed)
+
+
+const SHORTCUT_PANEL_MAX_PRESENTED_SHORTCUTS = 5
+@export var _shortcut_panel_presented_shortcuts := {}:  # Dictionary{int: String}
+	set(new_value):
+		if _shortcut_panel_presented_shortcuts != new_value:
+			_shortcut_panel_presented_shortcuts = new_value
+			_validate_shortcut_panel_presented_shortcuts()
```
syntaxerror247, line 493: I don't think this method is needed.

src/config_classes/SaveData.gd
```diff
@@ -32,58 +36,482 @@ func get_setting_default(setting: String) -> Variant:
...
+			emit_changed()
+
+# Remove invalid shortcuts.
+func _validate_shortcut_panel_presented_shortcuts() -> void:
+	for key in _shortcut_panel_presented_shortcuts:
+		if key < 0 or key >= SHORTCUT_PANEL_MAX_PRESENTED_SHORTCUTS or\
+		not _shortcut_panel_presented_shortcuts[key] in ShortcutUtils.get_all_shortcuts():
+			_shortcut_panel_presented_shortcuts.erase(key)
```
syntaxerror247, line 501: I don't think this method is needed.

src/ui_parts/shortcut_panel_config.gd
```diff
@@ -11,9 +11,12 @@ func _ready() -> void:
 	%LockPanelPosition.text = Translator.translate("Lock panel position")
 	%CloseButton.text = Translator.translate("Close")
 	%CloseButton.grab_focus()
-	%VerticalPanel.button_pressed = Configs.savedata.vertical_panel
-	%LockPanelPosition.button_pressed = Configs.savedata.lock_panel_position
 	setup_shortcut_slots()
+	# TODO Implement these properly when more configs are implemented.
+	%VerticalPanel.queue_free()
+	%LockPanelPosition.queue_free()
```
syntaxerror247, line 17: Why remove these buttons? I intentionally kept them to fill the empty space; otherwise, leaving it blank would look odd. Additionally, they are disabled, so users won't be able to interact with them anyway.

```suggestion
```

src/config_classes/SaveData.gd
MewPurPur: It's needed. Savefile loading should assume any input could happen. So any valid-looking dictionary would be accepted, but some wouldn't be contorming.

If a user goes in the savefile and changes "undo" to "qwerty", right now, your shortcuts panel would crash. Or more plausibly, if alpha 8 implements a new "qwerty" shortcut, the savefile of a user has it, and then they open their old alpha 7 build.

This practice of validation is going to be implemented for every resource going forward.

syntaxerror247: Okay! I didn't consider the possibility of users manually modifying the savefile, as they don't have access to the users:// directory on Android. However, this makes sense for the Desktop build.

src/ui_parts/shortcut_panel_config.gd
MewPurPur: Because people would think something is wrong with the app because they can't use the buttons. I want to generally disband them until we settle on an implementation. (For example I think we should have max column count instead of just horizontal/vertical.

src/config_classes/SaveData.gd
MewPurPur: Backwards compatibility is relevant for Android too.

PR #1049, opened 2025-01-05
Shortcut panel tweaks and layouts

MewPurPur:
Tweaks some UI and logic. Implements shortcut panel layouts. Implements a way to have maps between dropdown values and the texts (ideally I'd like to make a more thorough rework, but for now this will do) which is useful for the shortcut panel config. Fixes issue where shortcut panel can be dragged off-screen. Increases max shortcuts from 5 to 6 and changes the logic to rely on this number when necessary. Some small miscellanous fixes, optimizations, and additions.


syntaxerror247:
I haven't did the code review yet, these are the things I noticed while testing changes locally.

I really like that shortcuts in Panel are updated live as I add them in the config. However, there are a few aspects I'm uncertain about.

- Why handle icon is stretched vertically? It looks odd. I think it would be better to keep some space on its left, top and bottom.
- when i set a shortcut in config, dropdown's size decreases due to clear button. Can you fix that? It looks odd to reduce it's size while there is enough space.
- Also, horizontal with two strips, rearranges the buttons. My idea was to just add a hidden second row after first 5 are filled, panel will have a extra button to show that row and this would be for both vertical and horizontal.
- Also, panel is still getting dragged off-screen. It seems to work only on top and left edge.

MewPurPur:
Addressed the first two points. No idea about the fourth one, and I have different ideas about the third one.

PR #1056, opened 2025-01-07
Dutch translation patch

jas31415:
Proofread and patched up the Dutch translation .po

### DIFF
--- a/translations/nl.po
+++ b/translations/nl.po
@@ -13,19 +13,19 @@ msgstr ""
 "X-Generator: Poedit 3.4.4\n"
 
 msgid "translation-credits"
-msgstr "Racer911dash1"
+msgstr "Racer911dash1, jas31415"
 
 #: src/autoload/HandlerGUI.gd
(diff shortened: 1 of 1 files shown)


MewPurPur:
You should remove the fuzzy markers. Those are basically a way to keep track of a translation if it changes by just a bit, but they aren't used as literal translations.

translations/nl.po
```diff
@@ -631,22 +627,18 @@ msgid "Paste XML"
...
 
 #: src/ui_parts/settings_menu.gd
-#, fuzzy
 msgid "New palette"
 msgstr "Nieuw element"
 
 #: src/ui_parts/settings_menu.gd
-#, fuzzy
 msgid "New palette from XML"
 msgstr "Nieuw element"
```
MewPurPur, line 635: It seems like these had their fuzzy marker removed, even though they weren't correct translations.

MewPurPur:
Thanks!

PR #1075, opened 2025-01-12
Tweak Android shortcut logic

MewPurPur:
Something suboptimal I found while trying to fix something else.


MewPurPur:
Doesn't work

PR #1081, opened 2025-01-14
Move to Godot 4.4.beta1

MewPurPur:
Welp this was more cursed than I expected...

– All the folders are now littered with .uid files that are useless to humans. Yay!
\+ I've typed all the dictionaries. I typed the const stuff too, because while automatic typing can establish that they are arrays or dictionaries, it can't establish the exact types.
– ColorPalette is now an official Godot class, so I renamed our class to just Palette. Backwards compatibility of palettes is gonna be broken again.
– The new String.num() was broken in pretty much all instances, so I had to make a new stringy function in Utils.
\+ Performance is massively buffed.

And not as technical of an issue, but as the lead developer, I'm enjoying my workflows less here. I might do some Godot PRs to solve my grievances.

Smaller things:

- Physics servers are now set to dummy
- Added new hovered+selected style to theming (i.e., in the built-in file dialog)
- Tooltips are no longer blurry
- Resolved a couple of TODOs


MewPurPur:
Merging in spite of the failed Android workflow.

PR #1085, opened 2025-01-18
Fix unnecessary SVG.changed emitting

MewPurPur:
Tweaks the logic for when SVG.changed is emitted. Fixes issue where UndoRedo isn't created when editing an id attribute.


MewPurPur:
Oops, something here got quacked up, I'll have to redo this PR.

PR #1114, opened 2025-02-02
Strip unnecessary networking libraries

MewPurPur:
I hope this will improve build size :3c


MewPurPur:
Looks like I don't understand something about these config options.

PR #1174, opened 2025-02-22
Rework the UI scale setting

MewPurPur:
I've removed enum dropdowns, replaced with the help of the dropdown value-text maps.

The current UI scale logic is now something like a fallback, with DPI once again used for the main guess for auto scale.

The Auto UI scale setting is removed, and the main one is now an enum. Only the allowed scales are shown.

A "Max" option is also available to increase the UI scale as much as possible.


MewPurPur:
Reverted back to the old auto scale logic, but with a web-specific branch using https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio

PR #1202, opened 2025-03-01
More static typing and micro-optimization

MewPurPur:
Also improves code consistency a bit. Does with for loops that's kinda sucky, but more optimal and strictly typed.


MewPurPur:
The 2nd option is a bit slower (0.1-0.2ms per thousand runs). I'll revert back to it if any kind of GDScript optimization of this stuff goes through.

PR #1227, opened 2025-03-11
Implement MMB to close tab

Kiisu-Master:


src/ui_parts/tab_bar.gd
```diff
@@ -160,13 +160,13 @@ func _gui_input(event: InputEvent) -> void:
 	
 	queue_redraw()
 	if event is InputEventMouseButton:
+		var hovered_idx := get_hovered_index()
```
MewPurPur, line 163: Can we only define this in the cases where it's used? Fetching the local mouse pos is surprisingly expensive.

MewPurPur:
In my opinion, this should be a setting, where the new behavior is the default but it can be turned off.

I used to have a laptop with a touchpad containing a "middle mouse key", and it was extraordinarily easy to try to focus a page using the left button, but then clicking the middle button by mistake and losing the page. I've turned the middle button off on my new laptop's touchpad (it was even more intrusive and I rarely ever used it), but it feels like a use case that wasn't really thought through.

So from these experiences, I think this needs to be configurable.

src/ui_parts/tab_bar.gd
MewPurPur: You're probably practically correct that it doesn't matter in this specific case. Not computing values that won't be used (especially when it's an expensive operation) is just best practice for optimization.

PR #1238, opened 2025-03-15
CI: Always build android release template

syntaxerror247:
Currently, the new Android release template is not automatically rebuilt when workflow changes occur. This can lead to issues where changes take effect, but the Android release template remains outdated. To prevent this, we have to manually clear the cache before each release -- a step that could be easily overlooked.

This PR ensures that the release template is also rebuilt whenever the workflow changes, eliminating the need to manually clear the cache before a release.


MewPurPur:
> Currently, the new Android release template is not automatically rebuilt when workflow changes occur.

Not sure I understand, doesn't this mean the condition should be changed to an "OR" rather removing the workflow-dispatch condition?

MewPurPur:
Thanks!

PR #1241, opened 2025-03-15
Update de.po

Swarkin:
I did this in the github web editor but it should work


MewPurPur:
These strings are commented out as they are no longer used. They are only kept for record-keeping, and in case new strings show up that use them or something similar, in which case the translations done there can still be restored or fuzzied.

Swarkin:
ohhh

PR #1242, opened 2025-03-16
added translation for tab bar settings

JinEnMok:
The new tab bar settings menu wasn't translated, so fixed that.


MewPurPur:
Thanks!

PR #1243, opened 2025-03-16
Fix CTRL + MMB zooming not keeping center at starting location

Kiisu-Master:
This was broken since the addition of fake mouse movement events


MewPurPur:
Thanks!

PR #1246, opened 2025-03-18
Fix tabbar error

MewPurPur:
ci is taken from the node and cleaned automatically, it's not like surfaces created through the RenderingServer. So cleaning it manually is wrong.


MewPurPur:
Kudos to @FlooferLand for finding this!

PR #1265, opened 2025-04-13
Add tab previews for unsaved files

FlooferLand:
Previously, files that were never saved on the disk didn't show any previews on the tab bar.

New:
image
_(not particularly useful for empty files)_


MewPurPur:
I think empty tabs having a preview is ok.

Cons: Takes more space, people think it's a mistake

Pros: You see the dimensions of the empty SVG (but that'll almost never be useful)

PR #1288, opened 2025-04-25
Fix update checker

MewPurPur:
Attempt with CryptoKey instead.


Swarkin:
try both at the same time if it still doesn't work?

PR #1290, opened 2025-04-26
Fix update checker in exported builds

MewPurPur:
Let's try with both CryptoKey and X509Certificate enabled.


MewPurPur:
This didn't work, now I'm testing on Linux if the issue arises from the class disabling.

PR #1324, opened 2025-05-20
Implement use element with handles

### LINKED ISSUE #1323: Implement moving `<use>` elements with handles for convenient instancing
Most unimplemented svg features are perfectly usable from the code editor, but the one limitation I came across was that, even though instancing with `<use>` elements works, moving them is rather inconvenient because it is only possible from code.

So I propose an implementation of `<use>` that would allow moving the element with handles, even before a more feature complete implementation is considered. To complement it, I feel like it would be nice to have `id` show up as editable on every element too.

https://github.com/user-attachments/assets/a7c85e02-8a58-41f5-a91f-166cf2ab26b3

Jordyfel:
Fix #1323

Implements `<use>`, and it's `href` attribute. Currently missing an icon, but let's leave that after the implementation is agreed upon.

### DIFF
--- a/src/data_classes/Attribute.gd
+++ b/src/data_classes/Attribute.gd
@@ -4,6 +4,8 @@ class_name Attribute extends RefCounted
 
 signal value_changed
 
+enum NameValidityLevel {VALID, INVALID_XML_NAMETOKEN, INVALID}
+
 var name: String
 var _value: String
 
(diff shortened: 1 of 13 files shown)


src/data_classes/Attribute.gd
```diff
@@ -35,3 +37,24 @@ func _format(text: String, _formatter: Formatter) -> String:
 func _init(new_name: String, init_value := "") -> void:
 	name = new_name
 	set_value(init_value)
+	
+static func get_name_validity(id: String) -> NameValidityLevel:
```
Jordyfel, line 41: Moving this is a bit arbitrary, but it felt weird to have a sibling class also use it. Maybe it belongs in its own file

src/data_classes/AttributeHref.gd
```diff
@@ -0,0 +1,13 @@
...
+
+func set_value(new_value: String) -> void:
+	super(new_value if get_validity(new_value) != NameValidityLevel.INVALID else "")
+	
+	
+static func get_validity(id: String) -> NameValidityLevel:
+	if id.is_empty():
+		return NameValidityLevel.INVALID
+	
+	# Allow '#'
```
MewPurPur, line 12:
```suggestion
	# Allow '#'.
```

src/ui_parts/handles_manager.gd
```diff
@@ -858,7 +860,7 @@ func _on_handle_added() -> void:
 func create_element_context(precise_pos: PackedFloat64Array) -> ContextPopup:
 	var btn_array: Array[Button] = []
 	const CONST_ARR: PackedStringArray = ["path", "circle", "ellipse", "rect", "line",
-			"polygon", "polyline"]
+			"polygon", "polyline", "use"]
 	for shape in CONST_ARR:
```
MewPurPur, line 864: Don't add it here, the support for `<use>` in the GUI is too weak at the moment for it to be visible to everyone.

The `<stop>` element was also introduced for convenience one alpha before I reworked the core infrastructure to make gradients possible.

src/data_classes/Attribute.gd
MewPurPur: I'm fine with this.

src/ui_parts/handles_manager.gd
Jordyfel: So should it be removed from inspector right click and add element menus too?

MewPurPur:
By the way, how do you want to be credited in GodSVG's "About..." dialog? With your real name, or only your Github name?

Jordyfel:
In general pretty much no preference, as it is currently I wouldn't want to be the only one sticking out with a real name

src/ui_widgets/href_field.gd
```diff
@@ -0,0 +1,46 @@
+extends BetterLineEdit
```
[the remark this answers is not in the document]
MewPurPur: That's fine, I prefer it to be separate for now.

src/ui_parts/handles_manager.gd
MewPurPur: Yeah, make it impossible to add from the GUI. I guess I missed that. I will do a test run when I'm able to later.

MewPurPur:
I tried it out. It's far from a complete implementation, but since it's not exposed, it's just a nice little utility for anyone who needs it.

MewPurPur:
Thanks!

PR #1326, opened 2025-05-21
Implement constraining of control curve handles

### LINKED ISSUE #1325: With 2 consecutive beziers, allow constraining the moved control point to the line of the opposite control point
When making paths, I've found myself wanting to connect consecutive curves in a way that results in a "continuous curve", where there are no visible corners as one curve transitions to the other. To achieve this, the control points that connect to both sides of the curve connection should be on the same straight line. This is what `s` and `t` already achieve, but they also have the limitation that the distance from the control point to the connection cannot be edited independently.

In order to make such continuous curves more flexibly, it would be convenient to constrain the edited control point to the line formed between the connection and the other control point:

https://github.com/user-attachments/assets/646553ba-6a65-4907-82ad-785ff26e9db7

Jordyfel:
Fix #1325

Ended up implementing it as a namespaced (in an inner class) static pure function, so it can easily be moved.

### DIFF
--- a/src/ui_widgets/PathHandle.gd
+++ b/src/ui_widgets/PathHandle.gd
@@ -20,9 +20,12 @@ func set_pos(new_pos: PackedFloat64Array) -> void:
 	if precise_pos != new_pos:
 		var path_attribute: AttributePathdata = element.get_attribute(pathdata_name)
 		var cmd := path_attribute.get_command(command_index)
+		if Input.is_key_pressed(KEY_CTRL):
+			new_pos = LineConstraint.constrain(path_attribute, command_index, new_pos, x_param, y_param)
 		if cmd.relative:
 			new_pos[0] -= cmd.start_x
 			new_pos[1] -= cmd.start_y
+		
 		path_attribute.set_command_property(command_index, x_param, new_pos[0])
 		path_attribute.set_command_property(command_index, y_param, new_pos[1])
 		sync()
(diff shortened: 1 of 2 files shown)


MewPurPur:
This is a pretty clean implementation. I won't merge the PR for now, but I'll keep it for future reference.

MewPurPur:
I've not looked into this deeply yet, as you probably know the focus on the prerelease I'm planning this Sunday shifted a little, as I found some nasty issues I wanted to fix as soon as possible. But in any case, maintaining 64-bit precision is important.

MewPurPur:
I'll merge this for alpha 11, even though I want to rework the underlying systems soon. The only thing I want to suggest is, can the default keybind be made to be Ctrl?

src/ui_widgets/PathHandle.gd
```diff
@@ -20,9 +20,12 @@ func set_pos(new_pos: PackedFloat64Array) -> void:
 	if precise_pos != new_pos:
 		var path_attribute: AttributePathdata = element.get_attribute(pathdata_name)
 		var cmd := path_attribute.get_command(command_index)
+		if Input.is_physical_key_pressed(KEY_CTRL):
```
MewPurPur, line 23: Use is_key_pressed(), we're not using physical keys in GodSVG by convention (it's more a thing for games)

Jordyfel:
What are your thoughts on `Vector2.project()`, and implementing `Vector2` functions in general? Using it for the function should be more correct, though there will probably be no practical difference. The need to use vector math functions could come up again though, so it would be possible to either reimplement them with floats in a separate file use double precision builds of the engine.

MewPurPur:
I think we should implement 64-bit version of functions only when they are needed. I also think it's ok to implement a common function even if it's used only once for now.

I've thought about double-precision builds, but I deemed them more trouble than their worth. Utils64Bit came out of a few days of consideration for how to deal with 32-bit precision being insufficient.

Jordyfel:
This is what the code ends up looking like. Where should the vector2 functions go?

MewPurPur:
Utils64Bit.gd

PR #1329, opened 2025-05-22
Fix glitchy handle movement on android

### LINKED ISSUE #1055: Web: Staggering when draging PathHandle/PolyHandle
Caught in [web editor](https://www.godsvg.com/editor/) 

https://github.com/user-attachments/assets/00bac42e-d18b-47cc-b21f-4f7a256c1185

The point is a assigned a value the makes it go off screen, for a second before returning , and occurs numerous times as you drag a point.
seems to affect only polygon and polyline tags
edit: also path is affected

### LINKED ISSUE #1328: Android: Editing of paths with handles is glitchy
On android when moving handles for paths (maybe other things too), the handle will occasionally jump far and then come back to the correct place, which makes editing paths very unpleasant.

Debugging revealed that the problem is here:

Image

`Viewport.get_mouse_position()` sometimes returns 0.

Maybe its a godot bug with android, maybe it's an interaction with GodSVG's event throwing.

Jordyfel:
Fix #1328

*Production edit: Also fixes #1055*

### DIFF
--- a/src/autoload/HandlerGUI.gd
+++ b/src/autoload/HandlerGUI.gd
@@ -529,7 +529,8 @@ func throw_mouse_motion_event() -> void:
 	var window := get_window()
 	# Must multiply by the final transform because the InputEvent is not yet parsed.
 	var mouse_position = window.get_mouse_position()
-	# On android, this sometimes returns 0.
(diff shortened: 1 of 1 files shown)


MewPurPur:
Thanks!

PR #1346, opened 2025-06-03
Update translations for portuguese brazillian

Felipe-Sena:
Hi, this updates the Portuguese Brazillian translation strings and also fixes a few things I left as "needs work" a while back.


MewPurPur:
Thanks!

PR #1348, opened 2025-06-04
Add setting to limit FPS to any value

### LINKED ISSUE #1342: Add FPS limiter
Add settings, that will allow limiting how much GodSVG could render frames per second.
It would be especially nice if user could input FPS manually, not limited to pre-select list, like most games usually do.

Might be good idea to add "Apply" button when FPS number is changed, not when user changed number, to prevent issues when use could accidentally put 1 which will make app almost unresponsive.

Inputting 0 should behave like Godot editor does - unlimited FPS.

The idea here is:
1. Save battery, reduce GPU and CPU usage, and less heat produced (as result, fans will stay on lower RPM's, so less noise) by device on something, that might not need to be drawn 120 times per second.
2. For logical reasons - some might prefer to use program on lower FPS (I do, as example).

Kiisu-Master:
Closes #1342 

~~Currently there's a bug where the default value 0 doesn't show in the UI when opening the settings.~~(I tried to fix it by checking if the currently displayed string is empty)

The FPS limit doesn't apply until you press enter.

### DIFF
--- a/src/ui_parts/settings_menu.gd
+++ b/src/ui_parts/settings_menu.gd
@@ -286,6 +286,13 @@ func setup_content() -> void:
 			add_dropdown(Translator.translate("UI scale"), dropdown_values, dropdown_map)
 			add_advice(Translator.translate("Changes the scale factor for the interface."))
 			
+			current_setup_setting = "ui_max_fps"
+			add_number_dropdown(Translator.translate("Max FPS"),
+					[0, 30, 60, 120, 240], true, false,
+					SaveData.UI_MAX_FPS_MIN, SaveData.UI_MAX_FPS_MAX)
+			add_advice(Translator.translate(
+					"Limits the framerate of the user interface. If set to 0, the framerate will be unlimited"))
+			
 			# Disable mouse wrap if not available.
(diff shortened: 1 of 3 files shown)


src/ui_parts/settings_menu.gd
```diff
@@ -286,6 +286,13 @@ func setup_content() -> void:
 			add_dropdown(Translator.translate("UI scale"), dropdown_values, dropdown_map)
 			add_advice(Translator.translate("Changes the scale factor for the interface."))
 			
+			current_setup_setting = "ui_max_fps"
+			add_number_dropdown(Translator.translate("Max FPS"),
+					[0, 30, 60, 120, 240], true, false,
+					SaveData.UI_MAX_FPS_MIN, SaveData.UI_MAX_FPS_MAX)
+			add_advice(Translator.translate(
+					"Limits the framerate of the user interface. If set to 0, the framerate will be unlimited"))
```
MewPurPur, line 294: Needs punctuation at the end.

I would prefer if this is documented on the dropdown itself (e.g. the 0 option being "Unlimited" instead), but this is fine for now.

MewPurPur (commented):
I'll merge this for alpha 11.

PR #1350, opened 2025-06-04
Poedit view source workaround

Kiisu-Master:
Adds a colon at the end of file paths so poedit allows viewing the source files directly.
It wont jump to the place the string is used, but the source view window has searching capability with ctrl+f so its easy to find the string.


MewPurPur:
Thanks!

PR #1360, opened 2025-06-09
Add better exporting/saving for Chromium-based browsers

FlooferLand:
This uses the new web filesystem access APis, making GodSVG no longer save a new file to the downloads folder.

This instead prompts the user once to save a file and stores a handle to that file that we can later write to all we want.

There are some limitations; Mainly, the handle for each file has to be requested again any time the user loads up GodSVG again, as well as the first time the user saves after loading a brand new file, but this isn't problematic and it still beats saving as a new file every time.

This PR works on my machine; I've tested saving/loading files, multiple tabs, and exporting. I haven't tested this thoroughly however and I'd appreciate if someone would before merging _(Especially the XML colour palette saving? Not sure what that is about)_

See also: https://developer.chrome.com/docs/capabilities/web-apis/file-system-access

### DIFF
--- a/src/config_classes/TabData.gd
+++ b/src/config_classes/TabData.gd
@@ -91,6 +91,18 @@ func setup_svg_text(new_text: String, fully_load := true) -> void:
 func get_svg_text() -> String:
 	return _svg_text
 
+## Gets the default name of the tab.
+## Should also be used to get the file name for save dialogs.
+func get_tab_name(styled := false) -> String:
+	var save_name := svg_file_path.get_file()
+	if save_name.is_empty():
+		var default := ""
+		if is_empty():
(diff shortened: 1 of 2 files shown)


FlooferLand:
Hmm.. This used to work right as I made this PR, but something broke regarding callback arguments.
No matter what I try to do, the `console.log("ARGS: ", args)` I have in `_web_on_file_saved` always prints `undefined` to the console and GDScript errors out when trying to access those values

FlooferLand:
For anyone interested, there were internal discussions and this PR is far too broken and flawed rn to be merged.
I think I'll pick it up again, since its so useful and improves the user experience loads, but rn its very messy and doesn't work right.

PR #1361, opened 2025-06-09
Disable cmdline loading on Android

syntaxerror247:
App currently tries to import the godot internal cmdline arg as an svg then shows a discarded dialog on every app launch.
image

This could be fixed using a different method instead of disabling it for Android. However, there's no native terminal on Android and I don't expect users to open the app using cmdline.


MewPurPur:
This was solved in a different way that addresses more issues.

PR #1369, opened 2025-06-11
Disable wrapping for PO files

MewPurPur:
Should make it less error prone to update them without a tool.


MewPurPur:
Hmm... We'll see.

Issue #1315, opened 2025-05-14
Collapsable menues


Tsunami014:
When you make an svg with lots of things in it, the list view becomes quite crowded. I suggest having the list items all collapsable to declutter the view.

I suggest you would click on the title bar (![what it looks like](https://github.com/user-attachments/assets/8f7bebdf-5cfd-498c-a28e-11955ef153e6)) to collapse the element, so only the title bar is shown. The nested elements should also not be visible when the title bar is shown (or maybe that can be an option?)

MewPurPur:
The inspector definitely needs to be reworked. I'm not sure if the rework will make it less compact or more. In any case, I think something for seeing way less info would be great to have.

But it will take a while, as I'm working on alpha 9 at the moment, and after that I'll be focusing on reworking the handles system.

Tsunami014:
OK cool, was just one idea for the implementation anyway!

Also, for however you do it, an idea could also be to hide handles for collapsed (or whatever your alternative is) elements to not have them cluttering the viewport. (Again, just an idea; it really depends on the implementation)

Thanks so much for this awesome program, can't wait for the next release!

MewPurPur:
I started thinking about what the inspector should look like, evaluating pros and cons of different approaches. I'm pretty set now on a completely different implementation, which makes collapsing pretty unnecessary. It's too early to showcase, but, something like this:

<img width="415" height="492" alt="Image" src="https://github.com/user-attachments/assets/1e6b3ab4-1611-45c6-b64d-ee6308275436" />

I think this issue would become obsolete if I push this through.

Tsunami014:
I mean, it just becomes less of a problem; like if you had a very large number of elements or something it would be annoying

Or if you were working on a part of the picture and didn't want to have the other parts in the way?

But that layout is significantly easier to collapse, and does mean it is less necessary

PR #1381, opened 2025-06-26
Collapsible Inspector Elements

HuntJSparra:
(Leaving this in draft until I finish further testing and get a better feel for how the UI is)

**Description:** Initial implementation of #1315. Allows user to hide the attributes and children of an element in the inspector by clicking its title bar.
**Motivation:** Collapsing groups is very helpful when working on more complex files.

**Details:** All children of `main_container` in `element_frame`s are hidden if the new `Element.meta_collapsed` field is true for their `element`.

**Limitations:** What elements are collapsed is **NOT** saved, but I think this is an acceptable tradeoff for not having separate meta file. If there is demand for this in the future, then that could be implemented independently of this PR.

### DIFF
--- a/src/data_classes/Element.gd
+++ b/src/data_classes/Element.gd
@@ -9,6 +9,8 @@ signal descendant_attribute_changed(name: String)
 var _child_elements: Array[XNode]
 var _attributes: Dictionary[String, Attribute]
 
+var meta_collapsed: bool
+
 func _init() -> void:
 	attribute_changed.connect(_on_attribute_changed)
 	ancestor_attribute_changed.connect(_on_ancestor_attribute_changed)
--- a/src/ui_widgets/element_frame.gd
+++ b/src/ui_widgets/element_frame.gd
@@ -77,6 +77,9 @@ func _ready() -> void:
 		main_container.add_child(child_xnodes_container)
 		for xnode_editor in XNodeChildrenBuilder.create(element):
 			child_xnodes_container.add_child(xnode_editor)
+	
+	for child in main_container.get_children():
+		child.visible = not element.meta_collapsed
 
 func _exit_tree() -> void:
 	RenderingServer.free_rid(surface)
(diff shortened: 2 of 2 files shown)


HuntJSparra:
That makes sense to me! I need to rethink my approach anyways since it doesn't work when `sync_elements()` rebuilds the elements (which is very often).

MewPurPur:
That's ok! My current thoughts about this...

- I don't want to save metadata to the SVG ever.
- There are some app-wide settings, like "Show handles", that are in SaveData, even though they aren't accessed through the settings menu. These get serialized and persist through sessions and SVG changes.
- There can be tab-specific settings in the future, people have asked me to have per-tab zoom for example. These are also serialized, but they would probably not persist through SVG changes. This is where collapsed elements may make sense to save.
- State is only for very transient things, like selections (even that might go in TabData tbh), hover, and drop locations. These don't persist.

When the SVG text is changed, a lot of this data needs to be cleared, as we can't know what part of the text specifically was changed, and doing "smart" guessing I don't think is a good idea because while in some scenarios it will make things better for users, it will likely be bug-prone and less predictable to users.

I've not tested this implementation, but I would assume it's cleared when you switch between tabs, which yeah, isn't acceptable.

HuntJSparra:
I agree with all of that.

I've had the same thoughts about saving metadata (it goes against GodSVG's philosophy, so it's a no-go even if handling XML namespaces wasn't a pain) and "smart guessing" (it's too likely to go wrong when there are identical elements). The only option that leaves (and that I can think of) is minimizing what and when elements are rebuilt. The main causes of `sync_elements()` that I aware of right now are:
1. **Switching tabs:** As you said, it should be easy to cache the (session) metadata and restore it when tabs are switched. If the SVG has changed since the tab was last loaded, I think it's fine for the metadata to be discarded.
2. **Code Editor:** Similar to (1), I think it is acceptable to discard the metadata when the SVG text is directly changed. I suspect the people who heavily use this are not working in the inspector anyways.
3. **Undo/Redo:** This would have to be modified to work more like the drag-drop and not call `sync_elements()` (or minimize the syncing to the modified element or its new+old parents).

(1) is no work and (2) should be easy, but (3) is a hard problem. I'll probably look at it more on-and-off inspector rework, but those are my thoughts for now.

PR #1392, opened 2025-07-02
Fix `TabBar::_make_custom_tooltip()` not retrieving `hovered_tab` (#1391)

### LINKED ISSUE #1391: `TabBar::_make_custom_tooltip()` not retrieving `hovered_tab`
On rare occasions, `TabBar::_make_custom_tooltip()` will be unable to get the tab the mouse is currently over. In a release build, this has no noticeable effect. However, in the editor, this throws an error when attempting to get `hovered_tab.svg_file_path` (since `hovered_tab` is nil).

I suspect that the root cause of this is that there is at least one frame of delay between the call to `_get_tooltip()` and to `_make_custom_tooltip()`, during which the mouse can move off of the tab. I am unsure of this though.

**Ways to Reproduce:**
_This issue is intermittent and may require several attempts to reproduce_
1. Drag the window between monitors.
2. (Windows) Double-click to toggle if the window is maximized. This happens most often when your mouse is positioned so that it will immediately be over a tab after the resize.
3. Drag the window.

HuntJSparra:
Fixes #1391 

This PR refactors `TabBar::_make_custom_tooltip()` and `TabBar::_get_tooltip()` so that `_make_custom_tooltip()` does not need to lookup the hovered tab based on mouse position.

**Changes:**
1. `_get_tooltip()` now returns the index of the hovered tab.
2. `_make_custom_tooltip()` now uses the tab index from the tooltip to determine the hovered tab.

### DIFF
--- a/src/ui_parts/tab_bar.gd
+++ b/src/ui_parts/tab_bar.gd
@@ -421,7 +421,7 @@ func _get_tooltip(at_position: Vector2) -> String:
 	else:
 		# Return tab index as metadata so _make_custom_tooltip can determine the tab
 		# even if the mouse moves.
-		return str(hovered_tab_idx)
+		return String.num_int64(hovered_tab_idx)
 
 func _make_custom_tooltip(for_text: String) -> Object:
 	if not for_text.is_valid_int():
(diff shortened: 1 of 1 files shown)


MewPurPur:
Thanks!

MewPurPur:
How do you want to be credited in the "About" of GodSVG? Any real names, or do I just use the "HuntJSparra" one?

PR #1393, opened 2025-07-04
Improve dropdown widget usability on mobile

syntaxerror247:
If internal `line_edit` is disabled, clicking on it would open dropdown menu.


MewPurPur:
Thanks!

PR #1396, opened 2025-07-08
Add async rendering support

sockeye-d:
I've made it so that the SVG image gets rasterized on a separate thread and the handles get "rendered" to a bunch of easy to draw lists on a separate thread. This drastically improves responsiveness—I'm able to edit a 65 kB SVG file at reasonable framerates now.

I also added a Stopwatch class I made for https://github.com/sockeye-d/sled (with slight modifications) since I needed it for perf testing, though I can remove it.

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -384,6 +384,7 @@ locale/translations=PackedStringArray("res://translations/bg.po", "res://transla
 
 [rendering]
 
-renderer/rendering_method="mobile"
+renderer/rendering_method="gl_compatibility"
+renderer/rendering_method.mobile="gl_compatibility"
 textures/vram_compression/import_etc2_astc=true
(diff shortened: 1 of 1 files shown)


sockeye-d:
Oh also zooming is a little jank now, as far as I know there's no simple way to fix it but I'm also not super familiar with GodSVG's rendering internals. It may be possible to store the transform in the DrawCommandCache object, then transform the surface from the old transformation to the new transformation by scaling whatever's currently drawn rather but still drawing textures and stuff a frame out of date.

MewPurPur:
Also I just pushed a PR adding the new import settings, so that should be cleaned up too.

sockeye-d:
I think this is fundamentally the wrong approach to this, and I think with better caching and data structures it could get equivalent perceived performance without the jank and multithreading. As such, I'm going to close this PR

PR #1459, opened 2025-08-07
Fix About menu

syntaxerror247:
Fixes the issue on the author page where no author or donor was displayed due to the `app_info.toml` file not being exported.


MewPurPur:
Thanks!

PR #1472, opened 2025-08-11
filled in missing translations and checked fuzzy translations for Dutch

jas31415:
I checked for all empty (msgstr "") and fuzzy (#, fuzzy) translations. I filled those in and corrected the fuzzy ones too.
Another thing I did was check for some inconsistencies with wording, for words like "handle" and "viewport", although I didn't do a checkup on the consistency of every recurring word out there.

### DIFF
--- a/translations/nl.po
+++ b/translations/nl.po
@@ -76,9 +76,8 @@ msgid "Restore"
 msgstr "Herstel"
 
 #: src/autoload/State.gd:
-#, fuzzy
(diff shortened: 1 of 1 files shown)


MewPurPur:
Thanks! I did recently add comments explaining some tricky translations, like where "Handle" could be confused for a verb. For the record, those all start with "#.", as .po has specific syntax for different types of comments. Thanks for checking that out too.

PR #1473, opened 2025-08-12
Implement modular canvas

MewPurPur:
I hate working with viewport nodes...


syntaxerror247:
Viewport in settings menu looks good and per-tab viewport zoom/position also works as expected.

syntaxerror247:
[Screencast_20250814_143638.webm](https://github.com/user-attachments/assets/e9e38f53-6e99-4cb1-a119-fdd8bd959b96)

Handle doesn't work anymore. Trying to move the handles, moves the viewport instead.
I am testing on Fedora 42 Linux using artifact from this PR.

PR #1477, opened 2025-08-16
Cache inner markup when changing viewbox rect

sockeye-d:
As the title says, this makes the `SVGParser::root_cutout_to_markup` function accept another parameter that contains the cached inner XML. If the actual SVG content is changed, then the inner markup needs to be re-"rendered," but if it hasn't changed, then it can just be reused.


sockeye-d:
I added comments to the new things and renamed `inner` to `inner_markup`

MewPurPur:
Thanks!

MewPurPur:
Thanks!

PR #1497, opened 2025-09-09
Improve Chinese translation

williamchange:


MewPurPur:
Thanks! I was going to plead on social media for someone to volunteer Chinese translations, since Simplified Chinese is so important, yet less than two thirds complete at this point...

Just for future reference, I was also halfway into drawing this for a social media post P:

image

williamchange:
> Also, for the record, you didn't add your credits, is that intentional?

Ah I probably forgot

PR #1576, opened 2025-11-05
Fix wl-copy usage v2

sockeye-d:
This also simplifies the temporary file usage. Since wl-copy receives the data over stdin, no temp file needs to be created.


MewPurPur:
Thanks!

PR #1578, opened 2025-11-06
Add multiselect support to the toggle relative button

sockeye-d:
Now if you select more than one element it will toggle all of them. This also makes it so that clicking the button doesn't select the command.

(don't ask why I needed this, lmao)

[Screencast_20251106_023925.webm](https://github.com/user-attachments/assets/cdc5a0d8-805e-454b-a350-d5d2bf3514ec)

### DIFF
--- a/src/ui_widgets/pathdata_field.gd
+++ b/src/ui_widgets/pathdata_field.gd
@@ -141,7 +141,14 @@ func update_parameter(new_value: float, property: String, idx: int) -> void:
 	State.save_svg()
 
 func _on_relative_button_pressed() -> void:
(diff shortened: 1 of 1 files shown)


sockeye-d:
I might make it so that it only toggles them if the one you're hovering is one that's already selected, instead of the >1 behavior it is right now, and also that it sets the relative mode to whatever the opposite of the hovering command is instead of toggling it for all of them.

MewPurPur:
If you don't want to implement that right now, feel free to only add the toggling option you implemented in the right-click popup. Others can be added in the future.

PR #1587, opened 2026-01-28
Pebble Draw Command preview and export

sockeye-d:
The [Pebble Draw Command (PDC) format](https://developer.rebble.io/guides/app-resources/pdc-format/) is a vector format used by the Pebble smartwatches to draw scalable graphics. This PR adds export support for it, as well as extending the preview panel to look similar what the images would like like on real hardware. This includes quantization to 64 colors and snapping vertices to 1 or 1/16th pixels, depending on the precision mode.

The implementation was based on [svg2pdc](https://github.com/pebble-examples/cards-example/blob/master/tools/svg2pdc.py).

More information: https://developer.rebble.io/tutorials/advanced/vector-animations/#drawing-a-pdc-image, https://developer.rebble.io/guides/app-resources/pdc-format/


sockeye-d:
Should all be fixed now

sockeye-d:
Rebased!

MewPurPur:
I'd suggest to tune down the export menu's VBoxes' spacings slightly, just enough to still fit everything webp-related without shifting the layout. Also, add the new fields to the focus sequence in `_ready()`.

sockeye-d:
I could estimate the size of the image with the actual resolution by extrapolating the 
clamped image's bytes per pixel to the full image's number of pixels, probably. Then I could add an `" (estimated)"` suffix when that happens.

The two SVGs are unused, but they were for the PDC preview checkbox in the previews panel — if we plan on bringing that back, I'll still need them.

src/utils/SVGPathUtils.gd
```diff
@@ -0,0 +1,255 @@
...
+
+
+class AngleTolerancePathGenerator extends PathGenerator:
+	var angle_tolerance: float = 1.0
+	func generate_curve(from: Curve2D) -> PackedVector2Array:
+		return from.tessellate(6, angle_tolerance)
+
+
+class EvenLengthPathGenerator extends PathGenerator:
+	var length_tolerance: float = 0.5
```
MewPurPur: Use inferred typing

src/utils/Utils64Bit.gd
```diff
@@ -84,3 +84,11 @@ static func vector_project(a: PackedFloat64Array, b: PackedFloat64Array) -> Pack
...
 static func dot(a: PackedFloat64Array, b: PackedFloat64Array) -> float:
 	return a[0] * b[0] + a[1] * b[1]
+
+
+static func transform_vector_array_mult(transform: PackedFloat64Array, vectors: PackedVector2Array) -> PackedVector2Array:
+	var result: PackedVector2Array
+	result.resize(vectors.size())
+	for i in vectors.size():
+		result[i] = get_vector(transform_vector_mult(transform, [vectors[i].x, vectors[i].y]))
+	return result
```
MewPurPur: Add a documentation comment.

src/data_classes/PDCImage.gd
```diff
@@ -0,0 +1,577 @@
...
+			buf.encode_s16(buf_index + 0, encode_point(point.x))
+			buf.encode_s16(buf_index + 2, encode_point(point.y))
+		return buf
+	
+	func decode(sp: StreamPeer, points_size: int, path_open_or_radius: int) -> void:
+		is_path_open = path_open_or_radius == 1
+		for point_index in points_size:
+			var x := sp.get_16()
+			var y := sp.get_16()
+			points.append(Vector2(x, y))
```
MewPurPur: I'd say no need for intermediate variables.

MewPurPur (commented):
Other notes: The ThemeUtils changes should be reverted. They aren't utilized yet and they seem very hacky.

Edit: Also, maybe pull the whole export size estimation inside an if else block after checking if the format is SVG? Right now you have a bug where big SVGs will trigger the "approximate" label, and also have this weird match setup.

MewPurPur:
The export menu was reworked, which requires significant changes for the PR. But it should overall fit into the new systems way better than it did with the old ones.

MewPurPur:
*poke*

sockeye-d:
*deflates*

I unfortunately have other things I need to do 😭 I'll try to get back to this when I have time and motivation to

PR #1676, opened 2026-04-12
Update README.md

julientaq:
Give another option for the folks stuck in osx, who want to try it without having to turn off gatekeeper, in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,16 @@
 
 ## Downloads and links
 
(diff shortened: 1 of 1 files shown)


MewPurPur:
This is also too much info for a very particular use case not relevant to many people, so I'd like it to be put behind a summary, similar to the Android verification stuff.

julientaq:
done :)

MewPurPur:
Thanks!

PR #1681, opened 2026-04-23
Improve Android scaling, implement DDS, other fixes and improvements

### LINKED ISSUE #1678: Android UI scale issues
### Version

All/unknown

### Description

I have an  okay Android tablet, I went looking for Inkscape alternatives for Android and found GodSVG. Installed it and found it unusable.

Scaling seems rigid pixel based rather than physical (Inch per pixel). I need a magnifying glass to read/see anything useful in the UI and as such I can't include the version number in this report.

Please test Android builds on Android devices of varying resolutions before releasing Android builds.

Don't just build with a profile in Godot and assume everything "just works".

Happy to contribute, been doing Godot for years, even building projects that play nice between desktop and mobile, but I can't see enough to know where to start.

MewPurPur:
Fixes #1678, or at least tries to.

### DIFF
--- a/src/autoload/HandlerGUI.gd
+++ b/src/autoload/HandlerGUI.gd
@@ -505,21 +505,20 @@ func is_node_on_top_menu_or_popup(node: Node) -> bool:
 func get_window_default_size() -> Vector2i:
 	return Vector2i(ProjectSettings.get_setting("display/window/size/viewport_width"), ProjectSettings.get_setting("display/window/size/viewport_height"))
 
-func get_usable_rect() -> Vector2i:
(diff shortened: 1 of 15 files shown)


MewPurPur:
Android seems pretty broken on my device right now, I'll look into it after work.

MewPurPur:
I'm pretty sure this should fix Android, so I'm closing the mentioned issue.

I included a lot more fixes and improvements in the PR, as I hanged on it while being unable to test why Android is glitched out. But I confirmed it's not a regression from this PR, and that new way auto scale is decided works at least on my phone.

PR #1694, opened 2026-05-06
Switch to immersive mode on Android

syntaxerror247:
fixes: #1645


MewPurPur:
I think it looks good. I'll keep it around for now, but taking the obscured area by corner radii is still necessary. Are corner radii a unique case or could there be other quirks obscuring parts of the UI?

I think a subtle gradient towards black on the side with the big margin would help make it feel more natural.

syntaxerror247:
> Are corner radii a unique case or could there be other quirks obscuring parts of the UI?

I think corner radii is the only remaining case. I'll look into it tommorrow, to update the logic to consider corner radii as well.

MewPurPur:
*poke*

MewPurPur:
*poke*

Alpha 16 is planned for the second half of this month, just seeing if we can get this in.

syntaxerror247:
Sorry for the long delay. It took me about a month to get back to this, as life has been quite busy. When you have a chance, please test it and let me know if it looks good to you

MewPurPur:
The list of disabled classes currently includes JavaClass, JavaClassWrapper, and JavaObject, so this wouldn't build correctly.

MewPurPur:
Thanks!

MewPurPur:
Mmh, this was a hasty merge. Only a few critical UI elements need to steer away from corners, not the whole application area

PR #1698, opened 2026-06-04
metainfo: support tablet/stylus and replace deprecated developer_name

adil192:
- Replaces the deprecated [`<developer_name/>`](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer_name) with `<developer/>`.
- Adds graphics tablet to the supported input methods. (I don't think GodSVG uses pressure/tilt values currently but that's not an issue.) https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-control


MewPurPur:
Thanks!

Issue #1598, opened 2026-02-08
Publish on Flathub for Linux


adil192:
### Feature Description

I would love GodSVG to be packaged as a flatpak so that we have better desktop integration (like being added to the apps list) and so that it appears in Linux app stores (that was the first place I looked for a code-focused SVG editor).

### Implementation

I've prepared everything needed to publish to Flathub:
- Draft PR: https://github.com/adil192/flathub/pull/2.

I added a github actions workflow to update the build options, godot version, etc from your repo, so manual maintenance largely isn't needed.

### Your role

If you're on board, we can make you a maintainer of the Flathub repo.

And it'd be great if you could do these tasks:
- Upload a file to `https://godsvg.com/.well-known/org.flathub.VerifiedApps.txt` to [verify the app](https://docs.flathub.org/docs/for-app-authors/verification).
- Let me know whether you want me to also be a maintainer of the Flathub repo

If you're not interested, I can still publish it on Flathub albeit as an unofficial package.

Please let me know your thoughts and if a verified package on Flathub is something you're interested in

MewPurPur:
I'm researching what this all means currently.

MewPurPur:
I'm working on the .desktop and .metainfo files. I'll commit them in a bit, but I'm still not sure if I'll be officially publishing on Flathub.

adil192:
@MewPurPur I see you're now building AppImages. I still think a flatpak would be better since (in most distros) AppImages aren't added to the apps list, don't register their mime times, don't update themselves, and won't show up in app stores. Proper AppImage integration usually has to be manually set up by the user which isn't the best experience imo. Flatpaks don't have this friction for the user.

That aside, your current AppImage build isn't very portable since none of the system environment/libs are included in it. I would recommend using `linuxdeploy` instead of `appimagetool`. appimagetool's README says "in most cases you will be better off using one of the higher-level tools instead of using appimagetool directly". `linuxdeploy` automatically pulls in the required libs from your build environment.

`linuxdeploy` is quite easy to use. You can see an example in my app's CI here: https://github.com/saber-notes/saber/blob/main/.github/workflows/linux.yml. If you would like, I could make a PR for this?

MewPurPur:
Sure, a PR like that would most likely be accepted. For now I've heard some things about Flathub from some developers, especially regarding permissions, so I'm a little scared as it seems high maintenance. I think for the time being I don't want to be involved in maintaining it.

adil192:
[...]

PR #1699, opened 2026-06-04
ci: build appimage with linuxdeploy

adil192:
Removes the manual low-level build steps from appimagetool (https://github.com/MewPurPur/GodSVG/issues/1598#issuecomment-4623575651).

It turns out I was mistaken saying the previous AppImage wasn't portable. No new libs are deemed necessary by linuxdeploy, so the two should be equivalently portable.

Some other changes:
- Switched to the SVG version of the icon which is generally preferred.
- Added [update information](https://github.com/AppImage/AppImageSpec/blob/master/draft.md#update-information) which is read by tools like Gear Lever and AppImageUpdate. For this to work, you need to upload the `.AppImage` and `.AppImage.zsync` separately to each GitHub release.

This is how the AppImage looks in Gear Lever before and after this change:

Screenshot From 2026-06-04 21-40-07
Screenshot From 2026-06-04 23-08-17

### DIFF
--- a/.github/workflows/export-optimized.yml
+++ b/.github/workflows/export-optimized.yml
@@ -117,7 +117,7 @@ jobs:
       - name: Build AppImage
         run: ./linuxdeploy-x86_64.AppImage --appdir AppDir --output appimage
         env:
-          # `latest-all` includes pre-releases. Once MewPurPur leaves alpha, change it to `latest`.
+          # `latest-all` includes pre-releases. Once GodSVG leaves alpha, change it to `latest`.
           LDAI_UPDATE_INFORMATION: gh-releases-zsync|MewPurPur|GodSVG|latest-all|${{ env.PROJECT_NAME }}_v*-x86_64.AppImage.zsync
           LDAI_OUTPUT: ${{ env.FULL_NAME }}-x86_64.AppImage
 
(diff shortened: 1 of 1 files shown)


.github/workflows/export-optimized.yml
```diff
@@ -96,35 +96,36 @@ jobs:
...
+          sudo apt-get install appstream libfuse2
+
+      - name: Prepare AppDir for AppImage
+        run: |
+          install -Dm755 godsvg/build/${{ env.FULL_NAME }}.x86_64 AppDir/usr/bin/godsvg
+          install -Dm644 godsvg/no_export/distribution/com.godsvg.GodSVG.desktop -t AppDir/usr/share/applications/
+          install -Dm644 godsvg/no_export/distribution/com.godsvg.GodSVG.metainfo.xml -t AppDir/usr/share/metainfo/
+          install -Dm644 godsvg/assets/logos/icon.svg AppDir/usr/share/icons/hicolor/scalable/apps/godsvg.svg
 
       - name: Download AppImageTool
```
MewPurPur: This title no longer makes sense, right?

adil192: fixed :)

MewPurPur:
Thanks!

PR #1702, opened 2026-07-04
Improve Android corner/cutout handling

MewPurPur:


MewPurPur:
Will need a few more improvements down the line.

PR #1707, opened 2026-07-06
Revised Chinese(Simplified) strings

williamchange:
Removed the fuzzy markers since the translations are still correct.


MewPurPur:
Thanks!

PR #1735, opened 2026-07-23
Fix number field sliders causing frames to drag

FlooferLand:
Fixes a pesky bug where inspector element frames were being drag around when a `number_field_with_slider` was being interacted with via the slider.

https://github.com/user-attachments/assets/2e96ae37-3524-4f2d-924d-6174d3c4ea08

Usually bugs like these always happen because an input event isn't being swallowed somewhere where it should be.
In this case, two nodes were visibly trying to use events from the mouse at the same time, and adding `accept_event()` to a few cases of `InputEventMouseMotion` in `_on_slider_gui_input` fixed the bug for me.

### DIFF
--- a/src/ui_widgets/number_field_with_slider.gd
+++ b/src/ui_widgets/number_field_with_slider.gd
@@ -128,16 +128,17 @@ func _on_slider_gui_input(event: InputEvent) -> void:
 			slider_dragged = true
 			initial_slider_value = element.get_attribute_value(attribute_name)
 			set_num(get_slider_value_at_y(event.position.y))
-		accept_event()
+			accept_event()
 	else:
 		if Utils.is_event_drag(event):
 			set_num(get_slider_value_at_y(event.position.y))
+			accept_event()
 		elif Utils.is_event_drag_end(event):
 			slider_dragged = false
 			var final_slider_value := get_slider_value_at_y(event.position.y)
 			if initial_slider_value != element.get_attribute(attribute_name).num_to_text(final_slider_value):
 				set_num(final_slider_value, true)
-		accept_event()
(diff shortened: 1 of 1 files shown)


MewPurPur:
It's now flickery and UndoRedo no longer works. Breaking UndoRedo to fix this isn't acceptable, so I can't merge this as is.

FlooferLand:
Seems to be fixed now, my bad. The way I swallowed the event in `number_field_with_slider` was too broad and ended up swallowing events for undo/redo.
I couldn't replicate the flicker.
You should probably order your undo/redo singleton or whatever is responsible for it *before* normal nodes can get to it so this can't happen to begin with

MewPurPur:
Still don't find this to be an overall improvement. The flicker is still a problem. If you drag outside of the little field, the value stops updating - this also makes it hard to change the opacity to 1 or to 0. And after one drag, the problem reappears.

https://github.com/user-attachments/assets/36130b1b-23e5-4d74-9ce8-ba112043a9a3

> You should probably order your undo/redo singleton or whatever is responsible for it before normal nodes can get to it so this can't happen to begin with

I have no idea why it wasn't updating, just pointing out it was a problem. I would also think this shouldn't happen. I can help you investigate how it works, if you need help.

MewPurPur:
I think I know why UndoRedo was glitched before. As you drag a slider, the changes don't get saved, there's only one save when you release the slider, if the new value is different. Something about your initial changes must have interfered with the "set final value" mechanism, resulting in the slider not actually changing the SVG as it's saved, thus no UndoRedo.

MewPurPur:
Closed as FlooferLand asked in private.

PR #1740, opened 2026-07-26
Add On and Off strings

AlejandroMoc:
A small tweak to make the checkbox strings translatable.
In Spanish I've adapted them as "Yes" and "No" to use the same space.

image

### DIFF
--- a/src/ui_widgets/setting_frame.gd
+++ b/src/ui_widgets/setting_frame.gd
@@ -59,7 +59,7 @@ func set_optimizer_info(example_root: ElementRoot, optimizer: Optimizer, main_te
 func permanent_disable_checkbox(checkbox_state: bool) -> void:
 	disabled = true
 	widget.set_pressed_no_signal(checkbox_state)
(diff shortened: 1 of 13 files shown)


MewPurPur:
"Yes" and "No" is semantically wrong, the settings represent a state, not a question. What's the full translation for On/Off? If there's a common abbreviation, it can also be used.

AlejandroMoc:
In this context I think it's natural since some other programs and apps (like Minecraft) frequently use it, and it looks more natural than the abbreviations or the whole words, while still keeping the sematic sense in Spanish:

https://algomasquetraducir.com/como-traducir-on-off-espanol-software-juegos-aplicaciones/

However I have updated it to use the abbreviations "Act." and "Desact." since the full words "Activado" and "Desactivado" don't fit the menu.

MewPurPur:
Alright, I'll treat this as a UI/layout problem rather than a localization one. I'm trying something out right now, if it works out, I'll make a PR that credits you.

###### I think Minecraft's UI is quite bad

PR #1760, opened 2026-07-31
Fix palette config labels not updating when language changes

williamchange:
fix for these labels not updating when language changes

image

### DIFF
--- a/src/ui_widgets/palette_config.gd
+++ b/src/ui_widgets/palette_config.gd
@@ -42,7 +42,11 @@ func _ready() -> void:
 	name_edit.text_submitted.connect(_on_name_edit_text_submitted)
 	Configs.theme_changed.connect(sync_theming)
 	sync_theming()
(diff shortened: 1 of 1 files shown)


src/ui_widgets/palette_config.gd
```diff
@@ -42,7 +42,11 @@ func _ready() -> void:
 	name_edit.text_submitted.connect(_on_name_edit_text_submitted)
 	Configs.theme_changed.connect(sync_theming)
 	sync_theming()
+	Configs.language_changed.connect(sync_localization)
 
+func sync_localization() -> void:
+	set_label_text("")
```
MewPurPur, line 48: This is not right, it clears the palette title. To be fair, if a palette's name is empty, it also needs to update to the locale's version of "Unnamed", so that part needs to be synced too, but this isn't right.

williamchange: ok I will close this since I do not understand this part enough

PR #1792, opened 2026-08-11
Update translation credits in sl.po

anderlli0053:
Updated translation credits with contributor information.


MewPurPur:
Thanks!

PR #1811, opened 2026-08-19
Fix polyline simplification loop range and step

kdahal7:
In ElementPolyline.gd simplify(), range(1, list_size - 2) was using step 1 starting at index 1 instead of step 2 starting at index 2. This caused float coordinate swapping across (x, y) pairs and out-of-bounds array access on list[i + 3] during the final iteration.


MewPurPur:
Yeah, this is a misshap from my last PR. Thanks!

PR #1813, opened 2026-08-19
Add toggle for panning with the left mouse button.

### LINKED ISSUE #1809: Change or remove the use of left mouse button for panning
### Feature Description

When attempting to select a control point, I tend to miss by half an inch every now and then and consequently get punished with moving the canvas, which has became quite annoying to unintentionally perform frequently. Additionally, the middle mouse button is already used for panning, so it strikes me as strange that LMB without modifiers has two responsibilities instead of just one.
Perhaps this can be easily remedied by removing `MOUSE_BUTTON_MASK_LEFT` from the button mask in [canvas_navigation.gd](https://github.com/MewPurPur/GodSVG/blob/main/src/ui_widgets/canvas_navigation.gd)'s `_unhandled_input`?

If keeping LMB for panning is desired, then perhaps either of the following compromises would feel suitable:
* Add a toggle by replacing `MOUSE_BUTTON_MASK_LEFT` with some SaveData variable set to either the mask constant or to zero.
* Require an additional input. A precedent, which I have seen in other 2D art software like Krita, is to use Space + LMB for panning.
As far as I have seen, none of the default shortcuts use Space.

Gonderage:
Closes #1809. Enabled by default to preserve current experience.
Feel free to reorder the setting and rewrite the preview text.

### DIFF
--- a/src/config_classes/SaveData.gd
+++ b/src/config_classes/SaveData.gd
@@ -110,6 +110,7 @@ func get_setting_default(setting: String) -> Variant:
 		
 		# Other
 		"invert_zoom": return false
+		"pan_with_lmb": return true
 		"wraparound_panning": return false
 		"panning_speed": return 20
 		"use_ctrl_for_zoom": return true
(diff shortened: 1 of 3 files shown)


MewPurPur:
Thanks!

PR #1826, opened 2026-08-25
Add the Lolcat language

FlooferLand:
Adds the LOLCAT (Kingdom of Cats) language.

Silly language originating from early 2010s cat memes

Relatively easy to maintain since I made it a script that generates the language file from the normal English `en.po`

image

image

image

Somewhat based on the Minecraft Lolcat language that was removed at some point (https://gist.github.com/Darkhax/e5519fe36f0b51fcb1c8)

If anything breaks in the future it shouldn't matter much as the language itself is meant to essentially be broken English.

Replace strings shouldn't be affected since my `update_lolcat.gd` script tries to specifically replace words

### DIFF
--- a/no_export/scripts/update_lolcat.gd
+++ b/no_export/scripts/update_lolcat.gd
@@ -1,0 +1,106 @@
+# Run with Ctrl+Shift+X.
+@tool
+extends EditorScript
+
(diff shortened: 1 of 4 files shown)


MewPurPur:
Kthx!!

PR #1841, opened 2026-09-13
Migrate GodSVG to Godot 4.8 dev5

MewPurPur:
Removes some optimizations from disabled_classes (see: https://github.com/godotengine/godot/pull/119048)

Applies new optimizations - disabling RenderingDevice and disabling 2D with the new flag. Let's see if it works out.

There's a UI regression with the titled panels that's already addressed by https://github.com/godotengine/godot/issues/123392

Android exports no longer need to have their version updated separately.

Fixed bug where submenus didn't consider the direction they need to pop up in correctly.

Renames "spacious" formatting option to "verbose".

Removes some typing jank that has been fixed in Godot.

Tweaks point colinearity heuristics and changed some logic of how it's applied to fix a few issues.

Deprecates Utils.cubic_bezier_point() is Godot's bezier_interpolate() does the same, possibly better.

Removed some obsolete things.

Reworked the heuristics around grid snapping and axis constraints.

Fixed up various stuff that needed to be changed because of Godot updates.


MewPurPur:
GODSVG WEB SIZE INCREASED BY 3.14 MEGABYTES?

This has to be a bug, has to be bisected by Godot versions.

MewPurPur:
Windows failed but it's fine for now

